PR #25, opened 2016-05-18
Send UDP/123 requests to the local IP to localhost.

yomimono:
Proxy NTP requests to the NTP listener on localhost.  This allows the
guest to make NTP requests to its default gateway and get a reasonable
answer, as long as the host is answering ntp queries on localhost.

Signed-off-by: Mindy Preston <mindy.preston@docker.com>

### DIFF
--- a/src/hostnet/lib/slirp.ml
+++ b/src/hostnet/lib/slirp.ml
@@ -114,10 +114,11 @@ let connect x peer_ip local_ip =
                         Tcpip_stack.IPV4.writev (Tcpip_stack.ipv4 s) ethernet_ip_hdr [ reply ];
                       end else begin
                         let payload = Cstruct.sub udp Wire_structs.sizeof_udp (length - Wire_structs.sizeof_udp) in
+                        let for_us = Ipaddr.V4.compare dst local_ip == 0 in
                         (* We handle DNS on port 53 ourselves, see [listen_udpv4] above *)
                         (* ... but if it's going to an external IP then we treat it like all other
                            UDP and NAT it *)
-                        if Ipaddr.V4.compare dst local_ip != 0 then begin
+                        if (not for_us) then begin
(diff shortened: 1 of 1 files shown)


djs55:
Looks good to me!

src/hostnet/lib/slirp.ml
```diff
@@ -114,17 +114,27 @@ let connect x peer_ip local_ip =
                         Tcpip_stack.IPV4.writev (Tcpip_stack.ipv4 s) ethernet_ip_hdr [ reply ];
                       end else begin
                         let payload = Cstruct.sub udp Wire_structs.sizeof_udp (length - Wire_structs.sizeof_udp) in
+                        let for_us = Ipaddr.V4.compare dst local_ip == 0 in
```
samoht: use `=` instead of `==` here?

src/hostnet/lib/slirp.ml
```diff
@@ -114,17 +114,27 @@ let connect x peer_ip local_ip =
...
+                        if (not for_us) then begin
                           Log.debug (fun f -> f "UDP %s:%d -> %s:%d len %d"
                                        (Ipaddr.V4.to_string src) src_port
                                        (Ipaddr.V4.to_string dst) dst_port
                                        length
                                    );
                           let reply buf = Tcpip_stack.UDPV4.writev ~source_ip:dst ~source_port:dst_port ~dest_ip:src ~dest_port:src_port (Tcpip_stack.udpv4 s) [ buf ] in
                           Socket.Datagram.input ~reply ~src:(src, src_port) ~dst:(dst, dst_port) ~payload
+                        end
+                        else if for_us && dst_port == 123 then begin
```
dsheets: And `=` rather than `==` here? I think in this case it doesn't matter but the intent is confusing.

PR #33, opened 2016-05-24
Move mirage-net-unix version constraint to conflicts

ijc25:
opam is complaining:

[WARNING] At ~/kits.opam/packages/tcpip/tcpip.999/opam:50:9:
          Version constraint (>= 1.1.0) no longer allowed in optional dependency (ignored).
          Use the 'conflicts' field instead.

As suggested turn the >= 1.1.0 version constraint into a conflict with < 1.1.0.

Signed-off-by: Ian Campbell <ian.campbell@docker.com>

### DIFF
--- a/packages/tcpip.999/opam
+++ b/packages/tcpip.999/opam
@@ -49,6 +49,9 @@ depends: [
 ]
 depopts: [
   "mirage-xen"
-  "mirage-net-unix" {>= "1.1.0"}
+  "mirage-net-unix"
+]
+conflicts: [
+  "mirage-net-unix" {< "1.1.0"}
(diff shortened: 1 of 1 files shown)


dsheets:
LGTM

yomimono:
This looks good to me, and the AppVeyor test failure is unrelated.

samoht:
@djs55 and @yomimono can we make sure that this change go upstream as well?

yomimono:
@samoht upstream has already fixed this more generally; the depopt is just `mirage-unix` rather than both `mirage-unix` and `mirage-net-unix`.

Issue #24, opened 2016-05-18
Add build instructions


MagnusS:
There are currently no build instructions in the README.

samoht:
The build instructions are currently very involved ... I think the simplest path would be to extract the sequence of `pins` from `circle.yml` and put that in the Makefile. And `circle.yml` could just call it directly.

MagnusS:
Build instructions are now in the [README](https://github.com/moby/vpnkit/blob/master/README.md#building-on-unix) so closing this issue.

PR #35, opened 2016-05-25
Attempt to make the build easier

djs55:
The idea is to have 3 phases:

1. `circle.yml` and `appveyor.yml` install basic prerequisites like `opam`
2. `make depends` installs the library dependencies and sets up pins
3. `make` builds the component

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,11 @@ depends:
 	opam pin add -n hostnet src/hostnet -y
 	opam pin add -n osx-daemon src/osx-daemon -y
 	opam pin add -n osx-hyperkit src/osx-hyperkit -y
+ifeq ($(OS),Windows_NT)
+	opam pin add -n slirp src/com.docker.slirp.exe -y
+else
 	opam pin add -n slirp src/com.docker.slirp -y
+endif
 	opam depext -u slirp
 	opam install --deps-only slirp -y
 
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,30 @@
-.PHONY: com.docker.slirp.exe install uninstall
+.PHONY: com.docker.slirp.exe com.docker.slirp install uninstall
+
+TARGETS :=
+ifeq ($(OS),Windows_NT)
+	TARGETS += com.docker.slirp.exe
+else
+	TARGETS += com.docker.slirp
+endif
+
+all: $(TARGETS)
 
 EXEDIR=C:\projects\vpnkit
 
 com.docker.slirp:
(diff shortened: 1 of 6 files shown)


appveyor.yml
```diff
@@ -10,11 +10,15 @@ environment:
   TESTS: "false"
 
 install:
-  - appveyor DownloadFile https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/appveyor-opam.sh
+  - appveyor DownloadFile https://dl.dropboxusercontent.com/s/eo4igttab8ipyle/opam32.tar.xz
```
yomimono: We probably should find somewhere else to keep this.  I got some download errors when trying out this branch.

Makefile
```diff
@@ -1,13 +1,34 @@
...
 com.docker.slirp:
-	cd src/com.docker.slirp && \
-	oasis setup && \
-	./configure && \
-	make && \
-	cp _build/src/main.native com.docker.slirp
+	$(MAKE) -C src/com.docker.slirp
+
+depends:
+	opam repo list | grep com-docker-slirp || opam repo add com-docker-slirp .
```
yomimono: I'd really like to avoid invoking `opam repo add` if at all possible, since its effects are global rather than confined to a particular switch.

samoht: yes I fully agree.

djs55: Note that all this patch is doing is moving it from the .yml to the Makefile, in order to make the Makefile work. I'm not an opam expert so suggestions welcome!

yomimono: True, but I don't generally run the stuff in arbitrary `.yml` files in my development environment without having to copy them out to a terminal or otherwise look at them first, giving me an opportunity to say "uh, no thanks" to `opam repo add`.

I'd much prefer to explicitly pin the upstream dependencies to the branches called out in the `url`s provided by the repository files; that at least limits the potential chaos to the user's current branch.

djs55: The build at the moment is quite difficult, and perhaps we're stuck choosing between a complicated build, where many steps are needed (as @samoht noted in #24) and an easy / magical build that's too magical because of its side-effects. Longer term things should simplify as we remove the pins. In the meantime is there anything sensible we could do or should we leave it how it is? In practice I'm not going to be able to look at it for a few weeks so perhaps I'll close this and make space for someone to suggest something better.

yomimono: The PR really does give an easier build and my main issue is the `opam repo add` -- I'll submit a PR with pins instead on top of this branch.  Thanks @djs55 !

PR #36, opened 2016-05-26
Have Dns_forward operate on ipv4 and udpv4 modules, not an entire stack

yomimono:
Users will no longer have to construct additional modules and records which won't be used by Dns_forward.


djs55:
Reducing the dependence on the whole `Tcpip_stack` seems like a good idea to me!

djs55:
It looks like the CI builds failed for unrelated network errors. I've restarted the CircleCI one but I don't have access to the appveyor rebuild button-- I think the only way to trigger it is by pushing new commits :/

avsm:
belated lgtm

PR #37, opened 2016-05-26
Support port forwarding on Windows

djs55:
This PR tidies up a few signatures, in particular trying to use `V1_LWT.FLOW` in most places. Note that
- the port forwarding client needs to be able to call `shutdown_write` separately from `close`, so it uses the `Mirage_flow_s.SHUTDOWNABLE` signature
- both vmnet and the port forwarding protocol want to read into a fixed buffer, so we extend the signature with `read_into`. I think the `V1_LWT.FLOW` should be extended with this; on balance it was a mistake to always return an arbitrarily-sized fresh buffer. It worked for vchan but is inconvenient for almost everyone else.

When using Hyper-V sockets we layer a protocol on top of the raw reading and writing to implement `shutdown_write` since the sockets themselves don't support it. The protocol is implemented in [djs55/ocaml-hvsock] and is supposed to match [rneugeba/virtsock]. It's not quite working yet.


src/hostnet/lib/conn_lwt_unix.ml
```diff
@@ -1,8 +1,123 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ *)
```
samoht: this doesn't match with the toplevel LICENSE file :p

djs55:
This requires https://github.com/djs55/ocaml-hvsock/pull/13

djs55:
The required ocaml-hvsock changes are in master, and will be in the v0.6 release.

yomimono:
I probably won't be able to give this the attention it deserves until next week. :/

djs55:
I'll rebase to fix the conflict later today / tomorrow and then check if this works end-to-end. FWIW it worked exactly once before I went on holiday :) Let me know what you think. I think there's still a bit of confusion (in my mind anyway) over the shutdown/close behaviour in `V1_LWT.FLOW` -- hopefully I articulated that in one of my emails to [mirageos-devel](http://lists.xenproject.org/archives/html/mirageos-devel/2016-06/msg00009.html). I think that particular confusion will have to be resolved elsewhere and the results propagated here.

djs55:
OK I've rebased this and done some simple testing. I'd like to merge it today if possible, to give time for a bit more test and fix before thursday. @yomimono @avsm @samoht any thoughts? Does that seem sensible?

src/hostnet/lib/conn_lwt_unix.ml
```diff
@@ -1,6 +1,104 @@
...
+
+let shutdown_write flow =
+  try
+    Lwt_unix.shutdown flow.fd Unix.SHUTDOWN_SEND;
+    Lwt.return ()
+  with
+  | Unix.Unix_error(Unix.ENOTCONN, _, _) -> Lwt.return ()
+  | e ->
+    Log.err (fun f -> f "Socket_stack.TCPV4.shutdown_write: caught %s returning Eof" (Printexc.to_string e));
+    Lwt.return ()
```
avsm: The nit about `Lwt.return ()` -> `Lwt.return_unit` -- none of them are in perf critical calls though

avsm:
@djs55 lgtm

samoht:
I'm not sure I understand enough of the context to give a sensible ack here. Would be very great to improve mli/docs later so I can try to be more useful :p (but please do not block on my review to merge that PR)

djs55:
OK I'll bend the rules a little and merge -- if anyone spots anything egregious then feel free to create an issue and assign it to me and I'll fix it up.

PR #38, opened 2016-05-27
Fallback to secondary DNS servers upon retries

ijc25:
If the first upstream DNS server fails to respond then the guest will retry, today we would forward the retry to the same server, whereas on a native system it would try the next server in `resolv.conf`. This makes vpnkit more susceptible to errors due to a flakey upsteam server.

To resolve this keep track of in progress DNS requests and if we observe a reuse of a transaction id round-robin through the upstream servers instead of reusing the first.

### DIFF
--- a/src/hostnet/lib/dns_forward.ml
+++ b/src/hostnet/lib/dns_forward.ml
@@ -33,6 +33,30 @@ let tidstr_of_dns dns =
 
 module Make(Ip: V1_LWT.IPV4) (Udp:V1_LWT.UDPV4) (Resolv_conf: Sig.RESOLV_CONF) = struct
 
+type transaction = {
+  mutable resolvers : (Ipaddr.t * int) list;
+  mutable used_resolvers : (Ipaddr.t * int) list;
+  mutable last_use: float;
+}
+
+let table = Hashtbl.create 7
+
+let start_reaper () =
+  let rec loop () =
+    let open Lwt.Infix in
+    Lwt_unix.sleep 60.
+    >>= fun () ->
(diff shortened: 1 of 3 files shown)


yomimono:
If I'm reading this control flow correctly, changes to `/etc/resolv.conf` won't be noticed if they occur between retries for the same transaction ID, as `Resolv_conf.get ()` is only called in the unparseable-packet or non-retry cases.  I found this a bit surprising; is it intended?

ijc25:
Yes, it kept things simpler.

yomimono:
This PR is definitely an improvement on the previous behavior, and I think it's mergeable as is.  Thanks for the patch, @ijc25 !

src/hostnet/lib/dns_forward.ml
```diff
@@ -14,62 +14,137 @@ module OptionThread = struct
...
+    | (_, Some { Dns.Packet.id }) -> Hashtbl.remove table id in
+
+  Log.debug (fun f -> f "DNS[%s] %s:%d -> %s %s" (tidstr_of_dns dns) src_str src_port dst_str (string_of_dns dns));
+
+  match dns with
+  | (_, None) -> begin
+    Resolv_conf.get ()
+    >>= function
+    | r::_ -> Lwt.return (Some r)
+    | _ -> Lwt.return None
```
avsm: As a minor optimisation, you can use `Lwt.return_none` and `Lwt.return_some r` here which will avoid a fresh allocation of those boxes.

avsm:
LGTM

PR #40, opened 2016-06-02
com.docker.slirp: Bump nofile rlimit to 10240

ijc25:
This is the maximum on OSX, while the default is 256 which can be quite quickly
exhausted by parallel network traffic.

Also plumb through as a command line option.

Signed-off-by: Ian Campbell <ian.campbell@docker.com>

### DIFF
--- a/src/com.docker.slirp/src/main.ml
+++ b/src/com.docker.slirp/src/main.ml
@@ -45,10 +45,19 @@ let start_port_forwarding port_control_path vsock_path =
 
 module Slirp_stack = Slirp.Make(Vmnet.Make(Hostnet.Conn_lwt_unix))(Resolv_conf)
 
-let main_t socket_path port_control_path vsock_path db_path debug =
+let set_nofile nofile =
+  let open Sys_resource.Resource in
+  let soft = Limit.Limit nofile in
+  let (_, hard) = Sys_resource_unix.getrlimit NOFILE in
+  Log.info (fun f -> f "Setting soft fd limit to %d" nofile);
+  try Sys_resource_unix.setrlimit NOFILE ~soft ~hard with
+  | Errno.Error ex -> Log.warn (fun f -> f "setrlimit failed: %s" (Errno.string_of_error ex))
+
+let main_t socket_path port_control_path vsock_path db_path nofile debug =
   Osx_reporter.install ~stdout:debug;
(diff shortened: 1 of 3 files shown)


src/com.docker.slirp/src/main.ml
```diff
@@ -45,10 +45,20 @@ let start_port_forwarding port_control_path vsock_path =
...
 
-let main_t socket_path port_control_path vsock_path db_path debug =
+let set_nofile nofile =
+  let open Sys_resource.Resource in
+  let soft = Limit.Limit nofile in
+  let (_, hard) = Sys_resource_unix.getrlimit NOFILE in
+  Log.info (fun f -> f "Setting soft fd limit to %d" nofile);
+  try Sys_resource_unix.setrlimit NOFILE ~soft ~hard with
+  | Errno.Error ex -> Log.warn (fun f -> f "setrlimit failed: %s" (Errno.string_of_error ex))
+  | _ -> ()
```
yomimono: Is the arbitrary catch for other exceptions necessary here?

yomimono:
Aside from an exception-catching nitpick, this looks good to me.  Thanks!

src/com.docker.slirp/src/main.ml
ijc25: Bit of a brain fart that, in that I was trying to express the concept of "otherwise continue" in what is clear a nonsense fashion. I'll drop it.

src/com.docker.slirp/src/main.ml
```diff
@@ -45,10 +45,19 @@ let start_port_forwarding port_control_path vsock_path =
 
 module Slirp_stack = Slirp.Make(Vmnet.Make(Hostnet.Conn_lwt_unix))(Resolv_conf)
 
-let main_t socket_path port_control_path vsock_path db_path debug =
+let set_nofile nofile =
+  let open Sys_resource.Resource in
+  let soft = Limit.Limit nofile in
+  let (_, hard) = Sys_resource_unix.getrlimit NOFILE in
+  Log.info (fun f -> f "Setting soft fd limit to %d" nofile);
+  try Sys_resource_unix.setrlimit NOFILE ~soft ~hard with
```
samoht: do you mean `match [...]` here?

samoht:
LGTM

PR #46, opened 2016-06-08
proto-vmnet: camlp4 -> ppx

yomimono:
Signed-off-by: Mindy Preston <mindy.preston@docker.com>

Issue #27, opened 2016-05-20
Reduce the number of dependencies on forked packages.


yomimono:
vpnkit depends on custom versions of the following packages:
- charrua-core
- lwt
- mirage-types-lwt
- mirage-types
- tcpip

If possible, we should reduce or eliminate these by upstreaming our changes or changing our code to conform with upstream changes.

dsheets:
The `lwt` fork is likely not necessary for vpnkit but is tracked by ocsigen/lwt#215.

samoht:
Work in progress in #262

Issue #18, opened 2016-05-18
Include NTP server info in DHCP response


djs55:
Since we run our own DHCP server using the [charrua](https://github.com/haesbaert/charrua-core) library and can proxy NTP to the host, we should tell the VM what to do by including the host IP as the NTP server in the DHCP response. 

This involves:
- investigating whether the [charrua](https://github.com/haesbaert/charrua-core) supports this DHCP option and adding it upstream if not
- reconfiguring the DHCP server in vpnkit

yomimono:
Fixed by #47.

PR #47, opened 2016-06-08
Use charrua-core 0.3.

### LINKED ISSUE #21: hostnet's use of dhcp-server should not use parse
The DHCP server/library used by hostnet, charrua-core, has [an exposed configuration type](https://github.com/haesbaert/charrua-core/blob/master/lib/dhcp_server.mli#L38).  `hostnet` could construct the configuration directly, rather than constructing a string and then requesting the DHCP server attempt to parse it into a configuration.  This would likely help in implementation of #18 .

yomimono:
Fixes #21 and #18.  Removes a pin, addressing #27 .

Follow-on of #46 .


yomimono:
Human-tested, robot-approved. @ijc25 , this may be too late to help you, but perhaps not?

samoht:
Did you run `make commit` inside `opam/` to update the vendored packages?

src/hostnet/lib/tcpip_stack.ml
```diff
@@ -29,37 +28,47 @@ let rec smallest_prefix a_ip other_ips = function
...
+    hostname = "vpnkit"; (* it's us! *)
+    hosts = [ hyperkit ];
+    default_lease_time = Int32.of_int (60 * 60 * 2); (* 2 hours, from charrua defaults *)
+    max_lease_time = Int32.of_int (60 * 60 * 24) ; (* 24 hours, from charrua defaults *)
+    ip_addr = local_ip;
+    mac_addr = server_macaddr;
+    network = prefix;
+    range = (low_ip, high_ip);
+  } in
+  { peer_ip; local_ip; prefix; server_macaddr; client_macaddr; dhcp_configuration }
```
djs55: That's looking a lot nicer!

BTW do you know if charrua allows the dynamic range to be empty? I tried various things with the string and failed to make it work.

yomimono: Sadly, no; it explicitly disallows that case when parsing a string, and it's not expressible in the type signature of the configuration record.  I filed an issue there: see https://github.com/haesbaert/charrua-core/issues/27 .

yomimono:
@samoht: I completely missed that there was a Makefile in there, but I think I did the equivalent manually minus the invocation of `./update-upstream.sh`.

djs55:
@samoht is it ok to merge this as-is and then use `make commit` next time? Or will that break something?

samoht:
as long as CI is happy it should be ok :-) I can add commit hooks to make the CI fail if you think that's useful.

djs55:
Failing CI with a message which says "use `make commit`" would be handy!

PR #48, opened 2016-06-09
[do not merge] Simplify build by removing osx-daemon internal package

djs55:
This only had

- a dead function (`get_trace_dir`)
- trivial and application-specific logging setup
- a trivial `listen` wrapper

Since these really aren't useful anywhere else, inline them and remove the `osx-daemon` package to simplify the project.


djs55:
Talking with @samoht we're not sure what to do with the `osx-daemon` package: for example should it be inlined into datakit too or should it remain as a library exported from here and used by datakit?

On the one hand, having conventions would be nice. On the other hand the actual amount of code being shared is more than the boilerplate (_oasis, opam, Makefile) required to build it which is always a bad sign.

dsheets:
`s/more/less/`? Do we think there will be further functionality we might include in `osx-daemon` in the future?

ijc25:
The function `listen` (which I'm not sure I would call trivial ;-)) doesn't seem specific to either `osx` or `daemon`s (although it's more plausibly related to the latter). Perhaps there is some existing library/project which might be a better home than a new project? e.g. An existing collection of daemon or network helper functions maybe?

Maybe the logging setup stuff would make a nice convenience API addition to the underlying logging library?

djs55:
As part of the merging of `com.docker.slirp` and `com.docker.slirp.exe` there was a bit of refactoring  and now I don't think we actually use this code in this repository. I'll close this PR and then make a new PR which simply deletes it from here and make a PR to add it into a new home elsewhere.

PR #49, opened 2016-06-09
Send TCP addressed to us, to us rather than to someone else

djs55:
Currently if the peer sends TCP traffic to us on port 53 then we process it. If the peer sends TCP traffic to us on other ports then we try to make a connection on on the *outside* network to the IP from the *inside* network. This seems wrong because

-  the internal IP is arbitrary and only valid on the inside network, so we should only use it on the inside network
- the peer has deliberately directly addressed us, so we should treat it as a connection request to us and respond as such.

This PR redirects connections sent to the gateway IP (i.e. "bound" to the virtual interface inside vpnkit) to 127.0.0.1 on the host running vpnkit. In the context of Docker for Mac this would allow services to be run on the host and then accessed from a suitably privileged container (`--privileged --net=host`).

### DIFF
--- a/src/hostnet/lib/slirp.ml
+++ b/src/hostnet/lib/slirp.ml
@@ -149,7 +149,8 @@ let connect x peer_ip local_ip =
                     (Ipaddr.V4.to_string src_ip) src_port
                     (Ipaddr.V4.to_string dst_ip) dst_port in
                 Log.debug (fun f -> f "%s connecting" description);
-                ( if Ipaddr.V4.compare src_ip local_ip == 0 && src_port = 53 then begin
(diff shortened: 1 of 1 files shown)


yomimono:
Looks very good to me.

ijc25:
Does this have any impact on the behaviour added in #25 ("Send UDP/123 requests to the local IP to localhost")? Looks like it maybe makes that special case obsolete?

djs55:
This PR only changes the TCP path -- the UDP/123 thing isn't obsolete yet. I thought I'd test the waters with TCP first -- perhaps we should do both?

ijc25:
Ah, I failed to spot the TCP vs UDP thing, I think doing one at a time is fine.

PR #57, opened 2016-06-11
Run unit tests for packages in this repo

djs55:
This attempts to activate unit tests in the CI. It splits the dependency installation into 2 phases:

1. install other dependencies without running their unit tests: this is mainly for speed and just in case someone else ships a broken unit test
2. install our packages with unit tests enabled

The second patch is an artificial failure injected to make sure the CI does fail.


djs55:
@samoht: is there a better way to do this with opam? Is it a good idea to avoid running other people's unit tests or should I drop that part?

samoht:
I think that's good idea! To be maybe a bit more generic, we could install all the upstream packages first (by listing the directories in `$(OPAM_REPO)/packages/upstream`) but the current logic is fine as well.

samoht:
Maybe we want to call `make test` when building `com.docker.slirp.{,exe}` as well?

djs55:
It looks like `logs` has a `{test}` dependency on `mtime`, which opam wants to install even though it won't actually be used because the tests of `logs` won't be run.

```
The following dependencies couldn't be met:
  - slirp -> logs -> mtime
Your request can't be satisfied:
  - No package named mtime found.
```

djs55:
Probably easiest to include the test dependencies in the opam universe -- I'll try that.

samoht:
Hum yes, this is a know defect of opam which would be great to be fix for 2.0.0 (I've added it to the list) as I've hit that issue too many times. Not sure how to fix that properly, but maybe we should just add mtime in the repo for now on.

djs55:
I tweaked the `update-upstream.sh` to include the test dependencies -- it turned out there were a few more. It should be easy enough to remove the argument again when `opam` 2.0.0 comes out (and I won't have to remember why it was the package was made into an exception!)

djs55:
It looks like the artificial failure triggered this time. I'll remove it now.

djs55:
Thanks!

PR #58, opened 2016-06-12
Add port forwarding unit tests

djs55:
- check that we can connect, send and receive data forwarded to a local service
- measure the time taken to perform 10 sequential connections

In particular the timing test will only fail of the connection is outrageously slow: on my MBP 10 connections takes about 0.003s. It will only take more than 1s if there's a > 100ms delay on the average connection.

### DIFF
--- a/src/hostnet/lib_test/forwarding.ml
+++ b/src/hostnet/lib_test/forwarding.ml
@@ -75,7 +75,29 @@ module LocalServer = struct
   let accept { listening_socket } =
     Lwt_unix.accept listening_socket
     >>= fun (fd, _) ->
-    Lwt.return fd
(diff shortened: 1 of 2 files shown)


djs55:
I don't want to merge this until after #57 is merged and the tests re-run to verify they pass on both OSX and Windows. In particular the timing test might fail on Windows (!)

samoht:
#57 is merged, can you rebase this PR?

djs55:
@samoht @yomimono rebased and CI has passed-- happy to merge?

yomimono:
@djs55 looks good to me!

djs55:
Thanks!

PR #59, opened 2016-06-13
When binds include localhost, try to bind ::1 as well

djs55:
On some platforms (at least Windows) accessing ports via `localhost` is very slow because the address `::1` will be tried first and will be allowed to timeout. This patch generalises the port bindings a little, allowing a single user-requested binding to be mapped to a list of listening sockets. When the user requests an address `0.0.0.0` or `127.0.0.1` we perform a best-effort bind to `::1` also to speed up accesses on these IPv6 platforms.

### DIFF
--- a/src/com.docker.slirp/src/bind.ml
+++ b/src/com.docker.slirp/src/bind.ml
@@ -23,7 +23,7 @@ let request_privileged_port local_ip local_port sock_stream =
         begin match r with
         | `Ok fd ->
           Log.debug (fun f -> f "Received fd successfully");
-          Lwt.return (Result.Ok fd)
+          Lwt.return (Result.Ok [ fd ])
         | `Error (`Msg x) ->
           Log.err (fun f -> f "Error binding to %s:%d: %s" (Ipaddr.V4.to_string local_ip) local_port x);
           Lwt.return (Result.Error (`Msg x))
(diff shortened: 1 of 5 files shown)


rneugeba:
I think this approach is OK. It's in line with what Linux does when `net.ipv6.bindv6only` is set to `0` (default). See also: https://github.com/docker/docker/issues/2174

yomimono:
CircleCI build failure looks like an actual problem:

```
File "src/bind.ml", line 37, characters 7-51:
Error: This expression has type
         (Lwt_unix.file_descr list, [> `Msg of bytes ]) Result.result Lwt.t
       but an expression was expected of type
         (Lwt_unix.file_descr, [> `Msg of bytes ]) Result.result Lwt.t
       Type Lwt_unix.file_descr list is not compatible with type
         Lwt_unix.file_descr 
```

djs55:
oops, yes indeed -- I was too focused on Windows. Will fix!

justincormack:
Seems reasonable though we may need to tweak later, not sure if we should respect the Linux side settings for ipv4 only and so on later...

djs55:
There may also be an assertion failure which shouldn't happen but which did just happen in my tests. Investigating.

djs55:
Darn, I've managed to trigger the `system lacked sufficient buffer space` error -- there might be a leak.

djs55:
The resource leak was actually due to using an older version of hvsock. With the latest [hvsock.0.7](https://github.com/ocaml/opam-repository/pull/6745) it seems to be fast and no sign of a resource leak after 10k connections.

I think this one is ready, assuming the CI passes.

yomimono:
That sounds good to me. :+1:

PR #62, opened 2016-06-13
Use io-page.1.6.1

### LINKED ISSUE #61: don't use io-page.1.6.0
```
io-page.1.6.0] https://github.com/mirage/io-page/archive/v1.6.0.tar.gz downloaded
```
-- unfortunately this will crash on windows when the GC activates.

djs55:
This avoids a very early crash when the GC calls the wrong Windows `free` function.

Fixes #61


djs55:
Merging to fix a crash in the CI built binaries.

PR #64, opened 2016-06-16
Add an OCaml patch to fix Windows select

### LINKED ISSUE #63: Windows: network stack sometimes blocks for ~30s under high load
When running lots of concurrent connections through the stack, occasionally all I/O stalls. The TCP timers still fire which suggests that the process itself is fine; however the I/O "engine" in Lwt seems to stop.

djs55:
Fixes #63 

Reported upstream with test case: http://caml.inria.fr/mantis/view.php?id=7276

This is similar to OCaml "PR#5563: harden Unix.select against file descriptors above FD_SETSIZE"
except that we can still handle these larger lists using the more generic
(but more complicated) codepath.

The MSDN documentation for select states that "The variable FD_SETSIZE
determines the maximum number of descriptors in a set.":

  https://msdn.microsoft.com/en-gb/library/windows/desktop/ms740141(v=vs.85).aspx

If we have a list of fds which is too large to be used in select(),
fall back to the generic codepath which is used for select()ing on multiple
types of handles at once.

Note the default FD_SETSIZE is 64, so it's fairly easy to hit this limit.

Note this asks the CI to rebuild OCaml rather than use a pre-built binary
so build times will be hit until we create a new cached build.

Note: the patch to the compiler = the existing patch to the compiler we're already using plus this small change:

https://github.com/djs55/ocaml/commit/535b3a623d92c69a8ef7224d3bb5184f800f6f56


yomimono:
This looks all right to me.  Thanks for the detective work and the fix, @djs55.

opam/win32/compilers/4.02.3+mingw64.comp
```diff
@@ -0,0 +1,26 @@
+opam-version: "1"
+version: "4.02.3"
+src: "http://caml.inria.fr/pub/distrib/ocaml-4.02/ocaml-4.02.3.tar.xz"
+patches:[
+"https://github.com/djs55/ocaml/commit/e5ff15e7b028445bcd5da88f90874bcc1e4597ac.diff"
```
samoht: is it possible to add your 2 lines patch on top of fdopen's patch? It would be easier to maintain when we decide to switch to 4.03.

samoht: (the patch in that list are applied in sequence before the compiling is built ?)

djs55: I tried that at first but I got the syntax wrong and all the patches were skipped. Is the syntax:
```
patches: [
  "one"
  "two"
]
```?

Comment (samoht) [inline opam/win32/compilers/4.02.3+mingw64.comp]: Normally yes.

Comment (djs55) [inline opam/win32/compilers/4.02.3+mingw64.comp]: OK I'll try it again. I think my previous attempt was
```
patches: [
 [ "one" ]
 [ "two" ]
]
```

samoht:
LGTM

djs55:
OK I've done some tests with the appveyor artefact and it's definitely working a lot better. I did see 1 timeout out of 5 test runs but TBH that could be the remote server or my Internet connection -- we really need a more contained test. Perhaps something like `iperf`? Throughout my tests I could `ping` (which I couldn't before) which suggests to me that the event handling is much better.

samoht:
Can we adapt some bits of the mirage-tcpip regression tests? We are doing interesting in-memory virtual networking in there. /cc @MagnusS

yomimono:
Yes, we can! I'm working on this right now, as a matter of fact.

rneugeba:
@djs55 @yomimono: I also started working on some tests creating many connections.

Issue #66, opened 2016-06-20
Lots of connections causes `Unix.select` to fail with `EINVAL`


djs55:
On OSX we see fatal errors from `Lwt_main.run` like

```
proxy: internal error, uncaught exception:
       Unix.Unix_error(Unix.EINVAL, "select", "")
       Raised by primitive operation at file "src/unix/lwt_engine.ml", line 371, characters 26-60
       Called from file "src/unix/lwt_engine.ml", line 312, characters 8-39
       Called from file "src/unix/lwt_main.ml", line 41, characters 8-82
```
This appears to be related to [ocsigen/lwt#222]

djs55:
It looks like even libev relies internally on `select` on OS X:
```
(libev) select: Invalid argument
```
I think we should switch to [libuv](https://github.com/libuv/libuv) as Windows is a first-class target via IOCP, it's used by node.js and they seem to know [all the tricks](http://code.saghul.net/index.php/2016/05/24/libuv-internals-the-osx-select2-trick/). Apparently `libuv` started off as a wrapper around `libev` which also supported Windows via IOCP, but recently the `libev` dependency has been removed and it's now a standalone library.

There seem to be several incomplete bindings:

- https://github.com/trevorsummerssmith/ocaml-libuv (using Ctypes)
- https://github.com/fdopen/uwt (hand-written, tuned C code)

There's a discussion comparing the two here: https://www.reddit.com/r/ocaml/comments/2wop3h/ctypes_based_bindings_to_libuv/

It's probably worth starting with `uwt` initially since the author is the person who maintains the OCaml on Windows port we're using and hopefully the Windows support is therefore first-class.

PR #67, opened 2016-06-21
Backport mirage-tcpip "Use Logs library for logging"

djs55:
Also known as 879b0fc68d2d44f5c6317f033ac50ffcc3d4277f

The backport is here:

https://github.com/djs55/mirage-tcpip/commit/0da78c96ab359252e473cd3c9e039680c6fb462e

In particular this avoids `Console.log_s` which won't work if we're using
`libuv` (or anything other than `Lwt_main.run` and `Lwt_unix`)

Related to #66 


samoht:
LGTM

PR #68, opened 2016-06-21
Remove dead code ("Socket_stack")

djs55:
We're only using the code in `module Socket`.

I think the code in `module Socket` (which deals with connecting and producing `FLOW`s should probably be replaced by [mirage/ocaml-conduit])


djs55:
Note this also helps #66 by reducing the amount of code that needs to be ported to `libuv`.

samoht:
Why was that code needed initially? The current code still try to `connect` right, so not really dead-code?

PR #69, opened 2016-06-30
Use libuv for I/O

### LINKED ISSUE #66: Lots of connections causes `Unix.select` to fail with `EINVAL`

djs55:
The current `Lwt_unix` implementation has problems on both Mac and Windows, specifically:

- on Mac, both the built-in default `Lwt_engine` and the `libev` one will use `select`. Unfortunately even with `DARWIN_UNLIMITIED_SELECT` we still see `EINVAL` errors (#66). Even if this worked it would not perform well with lots of connections.
- on Windows, the `select` `FD_SETSIZE` is 64 and the OCaml `Unix.select` falls back to a much more complicated emulated code path. The API we should be using for scalable networking on Windows is IO Completion Ports

This PR changes the socket side of the stack to use `libuv`, the abstraction layer used by node.js which uses IOCP on Windows and `kqueue` on Mac.

This PR changes all the direct uses of `Lwt_unix.file_descr` into abstract types `Socket.Stream.Tcp.t, `Socket.Stream.Udp.t` etc. This higher-level interface can then be mapped onto `libuv` via [fdopen/uwt] as well as traditional `Lwt_unix.

A side-effect is that duplicated code was coalesced-- in particular there were several places where `read` and `write` functions had slightly different error handling.

Another side-effect is adding a `Channel` on top of the `FLOW` to avoid reading packet-by-packet. This has increased performance -- some benchmarks to follow.

Note that we must avoid using libraries which use `Lwt_unix` directly, as these calls will block. Happily most Mirage libraries are already pure in this sense.

The unit tests have been extended to run both the `Lwt_unix` and `Uwt` versions. I was able to run `nmap` tests in containers and observe a stable number of fds (~2000) during the test runs with `lsof`.

Fixes #66


src/com.docker.slirp.exe/src/main.ml
```diff
@@ -21,149 +21,76 @@ let default d = function None -> d | Some x -> x
 let ethernet_serviceid = "30D48B34-7D27-4B0B-AAAF-BBBED334DD59"
 let ports_serviceid = "0B95756A-9985-48AD-9470-78E060895BE7"
 
-(* These could be shared with datakit. Perhaps they should live in mirage/conduit? *)
+let hvsock_addr_of_uri ~default_serviceid uri =
```
avsm: Yes this can indeed live in Conduit as a custom registration

src/com.docker.slirp/src/main.ml
```diff
@@ -20,30 +20,50 @@ let or_failwith = function
...
+module Config = Active_config.Make(Host.Time)(Host.Sockets.Stream.Unix)
+
+let unix_listen path =
+  let startswith prefix x =
+    let prefix' = String.length prefix in
+    let x' = String.length x in
+    prefix' <= x' && (String.sub x 0 prefix' = prefix) in
+  if startswith "fd:" path then begin
+    let i = String.sub path 3 (String.length path - 3) in
+    let x = try int_of_string i with _ -> failwith (Printf.sprintf "Failed to parse command-line argument [%s]" path) in
```
avsm: should be an Lwt.fail?

djs55:
Regression tests are looking good -- let's try to include this in Docker for Mac/Windows soon.

yomimono:
Belated LGTM :)

djs55:
Thanks :) (sorry for the rush)

avsm:
lgtm

PR #71, opened 2016-07-04
Monitor the database for upstream DNS server overrides

djs55:
If someone writes the key

  com.docker.driver.amd64-linux/slirp/dns

with \n-separated lines of the form

  a.b.c.d
  a.b.c.d#e

then these will override the default host DNS servers. If the key is deleted
then the original behaviour is restored (i.e. on Mac read /etc/resolv.conv
and on Windows use the command-line or 127.0.0.1:53)

This means that, for example, a UI would be able to choose the upstream
DNS server without the VM noticing the change. It also lets us separate the
detection of upstream DNS logic (which might use lots of platform-specific
magic) from the DNS forwarder itself-- we can keep the bulk of this code
pure and isolate the platform specific part using the database for communication.

### DIFF
--- a/src/com.docker.slirp.exe/src/main.ml
+++ b/src/com.docker.slirp.exe/src/main.ml
@@ -113,7 +113,7 @@ let main_t socket_url port_control_url max_connections db_path dns pcap debug =
   );
   Printexc.record_backtrace true;
 
-  Resolv_conf.set_dns dns;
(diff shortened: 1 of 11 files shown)


yomimono:
I thought loudly but apparently forgot to say that this LGTM.

djs55:
Thanks -- I'll hold off merging this for a little while until I've got something which works fully end-to-end.

yomimono:
This will conflict with recent changes that ijc made in DNS.  @djs55 , can you rebase?

djs55:
Will do.

PR #72, opened 2016-07-06
hostnet: fix TCP DNS (required when UDP responses are truncated)

djs55:
TCP connections sent to our IP are normally forwarded to the host,
except for port 53 (DNS) which is forwarded to an upstream DNS
server.

Previously we checked whether a connection was "for us" by comparing
the IP destination to our IP. If a connection was for port 53 we would
then rewrite the IP destination but we failed to reconsider whether
this new IP was ours or not. This meant that we always forwarded TCP
on 53 to localhost by mistake.

This patch makes the "is connection for us" decision into a
function `for_us` and re-evaluates it when the destination IP may have
changed.

This fixes a bug introduced in a56356fd427870f7f3679b8dc9db6f403ea394db

### DIFF
--- a/src/hostnet/lib/slirp.ml
+++ b/src/hostnet/lib/slirp.ml
@@ -153,8 +153,8 @@ let connect x peer_ip local_ip =
             Tcpip_stack.listen_tcpv4_flow s ~on_flow_arrival:(
               fun ~src:(src_ip, src_port) ~dst:(dst_ip, dst_port) ->
 
-                let for_us = Ipaddr.V4.compare src_ip local_ip == 0 in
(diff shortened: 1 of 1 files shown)


yomimono:
This looks good to me.  Thanks @djs55 !

djs55:
Thanks! I've written an integration test for this elsewhere, but we could certainly do with a component test here. I wonder if the Mirage DNS client has a TCP mode?

yomimono:
Unfortunately `ocaml-dns` doesn't support `tc` or TCP mode, but we could take a packet capture from something that does and fake it.

Duzhiok:
restart your computer and reinstall it.

PR #74, opened 2016-07-11
Update to uwt.0.0.3

djs55:
This simplifies the build by removing our customised `libuv` and `uwt.0.0.2` forked package.

Assuming this builds, we need to check the resulting binaries are compiled correctly on Mac. In particular we must ensure to build `libuv` from source as part of the `uwt` build, rather than using `brew` to make sure the `MACOSX_DEPLOYMENT_TARGET` is correct.


djs55:
Wow, it built ok. It looks like `uwt` will now build a static binary rather than a dynamic one:
```
$ otool -L com.docker.slirp
com.docker.slirp:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)

$ otool -l com.docker.slirp | grep MACOS -2
    uuid F66F919A-C2BC-3175-8810-7D54237E2920
Load command 9
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.10
```
The resulting binary looks ok, and it passed my integration tests.

@yomimono @samoht @avsm -- look ok to you?

samoht:
LGTM

PR #75, opened 2016-07-11
Fix bind errors

djs55:
The first three patches upgrade `uwt` and are the same as #74. The last patch fixes the error messages when port forwards fail in "expected" ways, e.g.

- `EADDRINUSE` should be printed as `Bind for %s:%d failed: port is already allocated` i.e. the port is already in use by something else
- `EADDRNOTAVAIL` should be printed as `listen <tcp|udp> %s:%d: bind: cannot assign requested address` i.e. the IP does not exist on the host

The pretty-printed error messages should now be the same as those printed by regular Docker on Linux.

This should remove errors like the mysterious `errno 526` which comes from the Linux 9P client not understanding the error response received from the 9P server in this repo.

### DIFF
--- a/src/hostnet/lib/forward.ml
+++ b/src/hostnet/lib/forward.ml
@@ -305,30 +305,55 @@ let start_udp_proxy description vsock_path_var remote_port server =
 let start vsock_path_var t =
   match t.local with
   | `Tcp (local_ip, local_port)  ->
-    check_bind_allowed (Ipaddr.V4 local_ip)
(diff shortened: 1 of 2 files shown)


djs55:
@yomimono @samoht @avsm any objections to the error message cleanups? The goal here is to make the errors the same as emitted by regular docker on linux.

yomimono:
No objection at all.

djs55:
Thanks!

PR #76, opened 2016-07-12
Add --version

samoht:


src/com.docker.slirp/src/main.ml
```diff
@@ -188,7 +188,7 @@ let command =
          flows via userspace sockets"]
   in
   Term.(pure main $ socket $ port_control_path $ vsock_path $ db_path $ nofile $ pcap $ select $ debug),
-  Term.info "proxy" ~doc ~man
```
djs55: Thanks for fixing the name -- I keep forgetting to do it!

djs55:
LGTM (assuming the topkg stuff builds, I've never used that before)

yomimono:
> would be probably better to turn com.docker.slirp and com.docker.slirp.exe into proper opam package and distribute the archive that they generate.

I agree with this.

samoht:
I've pushed a new version where topkg do the watermarking (thx @dbuenzli for providing the workaround)

samoht:
Not sure this worked properly on windows:
```
make[1]: Entering directory '/cygdrive/c/projects/vpnkit/src/com.docker.slirp.exe'
ocaml pkg/pkg.ml build --vcs true --pinned true
pkg.ml: [WARNING] Watermarks: OPAM fields of opam: topkg: no such command. Did you install topkg-care ?
pkg.ml: [WARNING] file opam: opam field "maintainer" undefined or unsupported
pkg.ml: [WARNING] file opam: opam field "authors" undefined or unsupported
pkg.ml: [WARNING] file opam: opam field "homepage" undefined or unsupported
pkg.ml: [WARNING] file opam: opam field "bug-reports" undefined or unsupported
pkg.ml: [WARNING] file opam: opam field "doc" undefined or unsupported
pkg.ml: [WARNING] file opam: opam field "license" undefined or unsupported
pkg.ml: [WARNING] file opam: opam field "dev-repo" undefined or unsupporte
```

I'll update the version of topkg we are using and check if we have the correct version number in the binary.

dbuenzli:
> pkg.ml: [WARNING] Watermarks: OPAM fields of opam: topkg: no such command. Did you install topkg-care ?

Did you check this ?

samoht:
On windows:

```
/com.docker.slirp.exe --version
135d82f-dirty
WARNING: Io_page on Windows doesn't guarantee alignment
```

I guess it means that it works... I guess the `dirty` flag is because we watermark the source code, which changes the current checkout.

samoht:
(it's a bit scary that AppVeyor is building commits which doesn't exist in the Git tree, though)

ijc25:
@samoht looks like appveyor is building `refs/pull/76/merge` rather than `refs/pull/76/head`, i.e. it is testing the PR merged into master, that commit might appear in the tree if you hit the merge button? Or maybe not if master moved on between the test and merge.

I'd be more concerned with the `-dirty`.

dbuenzli:
> I guess it means that it works...

Yes however if you are using watermarks that are looked up in OPAM files like `PKG_ISSUES`, `PKG_HOMEPAGE` (see the [default watermarks](http://erratique.ch/software/topkg/doc/Topkg.Pkg.html#VALwatermarks)) these will be set to  `"undefined"` if `topkg-care` is not installed (this will change once `opam` v2 gives us a proper `opam query` command).

> I guess the dirty flag is because we watermark the source code, which changes the current checkout.

I don't think this is the case since `version` is determined before the watermarks are applied ([code](https://github.com/dbuenzli/topkg/blob/aed7432f8687d2a19f595655a85dec97b4e0f59a/src/topkg_pkg.ml#L168-L179)), I'd say the reason is different.

samoht:
@dbuenzli I can reproduce the `-dirty` locally so seems to be an issue with `topkg` (don't have much time to investigate today)

dbuenzli:
> @dbuenzli I can reproduce the -dirty locally so seems to be an issue with topkg (don't have much time to investigate today)

As usual it's difficult to provide insights without details always posts results with `TOPKG_VERBOSITY=debug`. In this case you'll see the `git describe` that is being performed and you can then test it in your context. Note that this is not something that `topkg` does itself, it is what is reported by a `git describe --dirty` so I'm not sure it's an issue with `topkg`, I rather think there is some real dirtyness in your setup.

samoht:
```
$ git describe --dirty --always
da89d65

$ ocaml pkg/pkg.ml build --vcs true --pinned true -v -v
pkg.ml: [INFO] topkg v0.7.6, running main
pkg.ml: [INFO] Build configuration:
               pkg-name: com.docker.slirp
               build-dir: _build
               vcs: true
               pinned: true
               tests:
               debug: true
pkg.ml: [EXEC] "'ocamlfind' 'ocamlc' '-config' 1>'/var/folders/9g/7vjfw6kn7k9bs721d_zjzn7h0000gn/T/pkg.ml9ff935topkg'"
pkg.ml: [EXEC] "'type' 'git' 1>'/dev/null' 2>'/dev/null'"
pkg.ml: [EXEC] "'git' 'rev-parse' '--git-dir' 1>'/var/folders/9g/7vjfw6kn7k9bs721d_zjzn7h0000gn/T/pkg.mlb40e87topkg' 2>'/dev/null'"
pkg.ml: [EXEC] "'git' '--git-dir' '/Users/thomas/git/vpnkit/.git' 'describe' '--always' '--dirty' 1>'/var/folders/9g/7vjfw6kn7k9bs721d_zjzn7h0000gn/T/pkg.mlfc5da5topkg'"
pkg.ml: [INFO] Watermark NAME = "com.docker.slirp"
pkg.ml: [INFO] Watermark VERSION = "da89d65-dirty"
pkg.ml: [INFO] Watermark VERSION_NUM = "da89d65-dirty"
pkg.ml: [EXEC] "'git' 'rev-parse' '--git-dir' 1>'/var/folders/9g/7vjfw6kn7k9bs721d_zjzn7h0000gn/T/pkg.ml2f2670topkg' 2>'/dev/null'"
pkg.ml: [EXEC] "'git' '--git-dir' '/Users/thomas/git/vpnkit/.git' 'rev-parse' 'HEAD' 1>'/var/folders/9g/7vjfw6kn7k9bs721d_zjzn7h0000gn/T/pkg.mld7493ftopkg'"
pkg.ml: [EXEC] "'git' '--git-dir' '/Users/thomas/git/vpnkit/.git' 'diff' '--quiet' 'HEAD' 2>'/dev/null'"
pkg.ml: [INFO] Watermark VCS_COMMIT_ID = "da89d659d2a317999dfaf23e5daf407569077bd2-dirty"
pkg.ml: [EXEC] "'type' 'topkg' 1>'/dev/null' 2>'/dev/null'"
pkg.ml: [EXEC] "'topkg' 'ipc' '--verbosity=debug' 'opam-fields' 'opam' 1>'/var/folders/9g/7vjfw6kn7k9bs721d_zjzn7h0000gn/T/pkg.ml333c02topkg'"
topkg: [INFO] topkg v0.7.6 running
topkg: [IPC] opam fields of opam with cwd /Users/thomas/git/vpnkit/src/com.docker.slirp
topkg: [INFO] Parsing OPAM file opam
pkg.ml: [INFO] Watermark PKG_MAINTAINER = "David Scott <dave.scott@docker.com>"
pkg.ml: [INFO] Watermark PKG_AUTHORS = "David Scott <dave.scott@docker.com>"
pkg.ml: [INFO] Watermark PKG_HOMEPAGE = "https://github.com/docker/vpnkit"
pkg.ml: [INFO] Watermark PKG_ISSUES = "https://github.com/docker/vpnkit/issues"
pkg.ml: [INFO] Watermark PKG_DOC = ""
pkg.ml: [INFO] Watermark PKG_LICENSE = ""
pkg.ml: [INFO] Watermark PKG_REPO = "https://github.com/docker/vpnkit.git"
pkg.ml: [INFO] Watermarked src/main.ml
pkg.ml: [EXEC] "'ocamlbuild' '-use-ocamlfind' '-classic-display' '-tag' 'debug' '-build-dir' '_build' 'opam' 'LICENSE.md' 'README.md' 'src/main.native'"

$ git describe --dirty --always
da89d65-dirty
```

dbuenzli:
After watermarking it is evident that this will be dirty. What is the result of: 

```
'git' '--git-dir' '/Users/thomas/git/vpnkit/.git' 'describe' '--always' '--dirty'
```
before you start (always use the full commands as executed).

samoht:
Hum, this is weird:

```
$ cd ~git/vpnkit
$ 'git' '--git-dir' '/Users/thomas/git/vpnkit/.git' 'describe' '--always' '--dirty'
da89d65
$ cd src/com.docker.slirp
$ 'git' '--git-dir' '/Users/thomas/git/vpnkit/.git' 'describe' '--always' '--dirty'
da89d65-dirty
```

dbuenzli:
Mon dieu.

samoht:
http://comments.gmane.org/gmane.comp.version-control.git/264768

ijc25:
You probably want `--work-dir` set to the root of the repo too...

ijc25:
Or maybe you want `-C` instead of both? (BTW, try replacing `describe` etc with `status` to see how dirty it things you are...)

samoht:
I confirm that `-C` works much better:

```
$ cd ~/git/vpnkit/src/com.docker.slirp
$ git  -C /Users/thomas/git/vpnkit/ describe --always --dirty
da89d65
```

dbuenzli:
Problem is that I'm not sure that's what we actually want in all contexts (e.g. you wouldn't want the watermarking process to affect all the work-tree, only the package's sub-tree).. This use of `topkg`'s VCS functionality is not conventional (that's the reason why the auto-detection of `vcs` returns `false` in your context). Better if we sit and have a chat to figure out what needs to be done in person, I'm here until sunday.

PR #77, opened 2016-07-12
Update to protocol-9p.0.7.0 and hvsock.0.8.1

djs55:
These are both released, but not present in the upstream package repositories yet.


djs55:
Since this changes only the dependency metadata and not the code, I'll go ahead and merge this.

PR #78, opened 2016-07-13
com.docker.slirp: remove unnecessary dependencies from opam

djs55:


djs55:
The true purpose of this is to trigger another merge commit built on appveyor since the last one failed for no reason and there is no rebuild button.

yomimono:
Not only does the change look great to me, but the motivation reveals nothing about any kind of problem with anything anywhere.

djs55:
`ctypes`failed to build because of a missing `libffi`. A previous working build of the same version looks ok. Perhaps CircleCI have removed `libffi` from their image?

PR #80, opened 2016-07-14
On startup log subcomponent versions

djs55:
To make understanding bug reports easier, the log output will now contain the versions of

- `uwt`
- `hvsock`
- `hostnet`

### DIFF
--- a/src/com.docker.slirp.exe/Makefile
+++ b/src/com.docker.slirp.exe/Makefile
@@ -3,8 +3,10 @@ build: src/depends.ml
 
 src/depends.ml: src/depends.ml.in
 	opam config subst src/depends.ml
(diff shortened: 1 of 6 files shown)


djs55:
@yomimono @samoht @avsm any objections to this wonderful PR? :)

samoht:
LGTM

@dbuenzli is it possible to watermak more complex strategies (instead of calling `opam config subst` + `sed`?)

dbuenzli:
Currently the best you can use is String (see [docs](http://erratique.ch/software/topkg/doc/Topkg.Pkg.html#TYPEwatermark)). 

Now the problem with this is that to define the value with the result of a program invocation, you'd need to call this on each `pkg.ml` invocation. 

I know I had at a certain point `Value of unit -> (string, R.msg) result)` in the design but I'm not sure why I eventually left it out. If you feel this can be useful file an issue.

PR #83, opened 2016-07-15
hvsock.0.9.0: add not-yet-released patch

djs55:
This prevents returning SOCKET_ERROR as a number of bytes in the case
that WSAGetLastError() returns 0 (which it has been observed to do with
Hyper-V sockets)


djs55:
This patch is for a dependency, not vpnit itself

PR #84, opened 2016-07-18
Add a secondary dns forwarder IP

ijc25:
The second commit is the interesting one. Its commit log:

    In #38 I attempted to address the problem of flakey upstream DNS by
    round-robinning DNS retries (detected via the DNS ID field), replicating the
    fallback behaviour a client which has been given multiple entries in
    resolv.conf despite only providing then with a single entry.
    
    However this turns out to have been insufficient because not all client side
    resolvers reuse the DNS ID on retry, so we do not see those retries as such and
    send them to the primary upstream DNS a second time.
    
    To address this create a second IP end point (on the same underlying device, so
    essentially an IP alias)  and expose it to the client (via DHCP) as a second
    DNS server which forwards to the secondary upstream DNS server and revert most
    of #38 such that for the existing local IP we always forward to the primary
    upstream DNS server (if the host only has a single DNS server configured then
    both end points will go to the that one).
    
    Third and subsequent host DNS servers are ignored.
    
    This has only been tested on Mac, but given an implementation of Resolv_conf
    which returns 2 (or more) DNS end points this should work for Windows too.
    
    This second IP address is bolted onto the side in what appears to be a slightly
    odd way. However I think it is consistent with the direction (AIUI) we intend
    to be moving which is to remove the use of `Tcpip_stack_direct` (since we
    aren't really a TCP/IP "endpoint") in favour of a lower level approach (based
    on the newer Marshal/Unmarshal interfaces in mirage-tcpip rather than even more
    open coding). Currently mirage-tcpip does not support aliases (second IPs) on a
    device, so going the route of creating a second full stack would have required
    an entire extra network device be exposed to the client.

### DIFF
--- a/src/hostnet/lib/dns_forward.ml
+++ b/src/hostnet/lib/dns_forward.ml
@@ -113,8 +113,8 @@ let input ~ip ~udp ~src ~dst ~src_port buf =
       Log.debug (fun f -> f "DNS[%s] %s:%d <- %s %s" (tidstr_of_dns dns) src_str src_port dst_str (string_of_dns (parse_dns buffer)));
       remove_tid dns;
       Udp.write ~source_port:53 ~dest_ip:src ~dest_port:src_port udp buffer in
-
-    Socket.Datagram.input ~reply ~src:(Ipaddr.V4 src, src_port) ~dst:(dst, dst_port) ~payload:buf
+    let userdesc = "DNS[" ^ (tidstr_of_dns dns) ^ "]" in
+    Socket.Datagram.input ~userdesc ~reply ~src:(Ipaddr.V4 src, src_port) ~dst:(dst, dst_port) ~payload:buf ()
   | None ->
     Log.err (fun f -> f "DNS[%s] No upstream DNS server configured: dropping request" (tidstr_of_dns dns));
     Lwt.return_unit
--- a/src/hostnet/lib/dns_forward.ml
+++ b/src/hostnet/lib/dns_forward.ml
@@ -26,30 +26,7 @@ let tidstr_of_dns dns =
 
 module Make(Ip: V1_LWT.IPV4) (Udp:V1_LWT.UDPV4) (Resolv_conf: Sig.RESOLV_CONF) (Socket: Sig.SOCKETS) (Time: V1_LWT.TIME) = struct
 
-type transaction = {
-  mutable resolvers : (Ipaddr.t * int) list;
-  mutable used_resolvers : (Ipaddr.t * int) list;
-  mutable last_use: float;
-}
-
-let table = Hashtbl.create 7
-
-let start_reaper () =
-  let rec loop () =
-    Time.sleep 60.
-    >>= fun () ->
-    let snapshot = Hashtbl.copy table in
-    let now = Unix.gettimeofday () in
-    Hashtbl.iter (fun k trec ->
-      if now -. trec.last_use > 60. then begin
-        Log.debug (fun f -> f "DNS[%04x] Expiring DNS trec" k);
-        Hashtbl.remove table k
-      end
-    ) snapshot;
-    loop () in
-  loop ()
-
-let input ~ip ~udp ~src ~dst ~src_port buf =
+let input ~secondary ~ip ~udp ~src ~dst ~src_port buf =
   if List.mem dst (Ip.get_ip ip) then begin
 
   let src_str = Ipaddr.V4.to_string src in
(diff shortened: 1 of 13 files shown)


src/hostnet/lib/slirp.ml
```diff
@@ -152,9 +156,9 @@ let connect x peer_ip local_ip =
...
             Tcpip_stack.listen_tcpv4_flow s ~on_flow_arrival:(
               fun ~src:(src_ip, src_port) ~dst:(dst_ip, dst_port) ->
-
                 let for_us src_ip = Ipaddr.V4.compare src_ip local_ip == 0 in
-                ( if for_us src_ip && src_port = 53 then begin
+                let for_dns src_ip = for_us src_ip || Ipaddr.V4.compare src_ip extra_dns_ip == 0 in
+                ( if for_dns src_ip && src_port = 53 then begin
                     Resolv_conf.get () (* re-read /etc/resolv.conf *)
                     >>= function
                     | (Ipaddr.V4 ip, port) :: _  -> Lwt.return (ip, port)
```
djs55: Should we send the TCP traffic sent to the `extra_dns_ip` to the secondary DNS server as well as the UDP?

djs55:
LGTM

src/hostnet/lib/slirp.ml
ijc25: Ah yes, I punted on that because the current code wasn't doing round-robin yet either and then GC'd the `XXX` I'd left in a fit of cleanliness...

I think what I'd want to do is refactor the server selection stuff which I added to `Dns_forward.input` into a new helper which could be called here too to select the upstream server to go to.

src/hostnet/lib/host_uwt.ml
```diff
@@ -82,11 +82,14 @@ module Sockets = struct
         loop () in
       loop ()
 
-    let input ~reply ~src:(src, src_port) ~dst:(dst, dst_port) ~payload =
+    let input ?desc ~reply ~src:(src, src_port) ~dst:(dst, dst_port) ~payload () =
       (if Hashtbl.mem table (src, src_port, dst, dst_port) then begin
           Lwt.return (Some (Hashtbl.find table (src, src_port, dst, dst_port)))
         end else begin
-         let description = Ipaddr.to_string src ^ ":" ^ (string_of_int src_port) ^ "-" ^ (Ipaddr.to_string dst) ^ ":" ^ (string_of_int dst_port) in
+         let base_description = Ipaddr.to_string src ^ ":" ^ (string_of_int src_port) ^ "-" ^ (Ipaddr.to_string dst) ^ ":" ^ (string_of_int dst_port) in
```
avsm: This does a large number of intermediate string allocations on the input path. Would be better to replace any >size 2 concats with:

```
let base_description =
  String.concat [ el1; el2; ...; eln ]
```

but also if it's not used in the main datapath outside of logging, put this allocation inside the call to `Log.debug` (a few lines below) so that it will only get evaluated if the logger is active.

```
Log.debug (fun f -> foo ())
```

will only evaluate `foo ()` if debug logging is active.

src/hostnet/lib/slirp.ml
```diff
@@ -124,17 +124,21 @@ let connect x peer_ip local_ip =
                       end else begin
                         let payload = Cstruct.sub udp Wire_structs.sizeof_udp (length - Wire_structs.sizeof_udp) in
                         let for_us = Ipaddr.V4.compare dst local_ip == 0 in
+                        let for_extra_dns = Ipaddr.V4.compare dst extra_dns_ip == 0 in
```
avsm: these `==` should just be `=`. 

`=` is structural equality and `==` is for physical (pointer) equality in OCaml.

Doesnt matter too much for integers (which are immediates) but better for code style to uniformly use structural equality.

src/hostnet/lib/host_uwt.ml
ijc25: `String.concat` -> Sure.

Moving to the log -> The description is actually used in the record too:
```
           end else begin
             let last_use = Unix.gettimeofday () in
             let flow = { description; fd; last_use; reply} in
```
and then used from a bunch different Log calls (half a dozen here and the rest elsewhere e.g. on teardown). I think the correct extension of what you suggest is to turn `description` from a `string` into a `() -> string` and have the various logging locations evaluate it lazily, is that right?

src/hostnet/lib/slirp.ml
ijc25: I copied the line above ;-). Will fix.

src/hostnet/lib/dns_forward.ml
```diff
@@ -26,95 +26,34 @@ let tidstr_of_dns dns =
...
-  end;
+  Resolv_conf.get ()
+
+  >>= ( function
+  | _::sec::_ when secondary -> Lwt.return_some ("secondary", sec)
+  | pri::_::_ when not secondary -> Lwt.return_some ("primary", pri)
+  (* These two could be collapsed, but for the desire to differentiate in the logging *)
+  | pri::_ when secondary -> Lwt.return_some ("secondary(wrapped)", pri)
+  | pri::_ when not secondary -> Lwt.return_some ("primary(sole)", pri)
+  | _ -> Lwt.return_none )
```
yomimono: Minor style thing: these parens shouldn't be required.

src/hostnet/lib/host_uwt.ml
yomimono: The pattern you might like to emulate is something like `pp_flow`, where you take the set of things you want to describe and print to the formatter passed, like in https://github.com/mirage/mirage-tcpip/blob/master/lib/tcp/pcb.ml#L60 (as a somewhat arbitrary example).

src/hostnet/lib/dns_forward.ml
ijc25: Without them the final `| _ -> Lwt.return_none` was also capturing the following bind, leading to type hilarity. @djs55 and @avsm advised that the parens were the normal way to solve that.

src/hostnet/lib/host_uwt.ml
ijc25: Spoke with @avsm and @yomimono IRL and we decided that refactoring the logging here could wait for another PR. I'll do the `String.concat` thing now though.

src/hostnet/lib/dns_forward.ml
yomimono: Ah, sorry, context noticing fail.

ijc25:
Force pushed an update:
* Uses `String.concat`
* Uses `=` instead of `==`
* Renames `?desc` to `?userdesc`

I didn't do anything wrt lazier logging or to round-robin the TCP DNS (which is hard to do here because apparently our secondary DNS server doesn't do TCP, but also because it is incompatible with the "bolted onto the side" way I've done things here)

yomimono:
This change looks good to me.

PR #86, opened 2016-07-18
com.docker.slirp: remove 9P log spam

djs55:
The 9P server used to control port forwarding emits some particularly
spammy and not terribly useful debug messages. This patch suppresses
these by setting the minimum level to `Info` for the 9P messages.


yomimono:
Seems like a sensible change to me.

PR #87, opened 2016-07-18
Allow the max concurrent forwarded connections to be configured

djs55:
Every forwarded connection consumes resources:

- the accepted connection on the host
- the `virtio-vsock` or `AF_HYPERV` forwarded connection to the VM
- the accepted connection in the VM
- the forwarded connection in the VM

We've already bumped the maximum number of fds in various places, so now the bottleneck is the `virtio-vsock` or `AF_HYPERV` connection to the VM. We would like to avoid hitting this limit because it will cause the docker socket forward to fail too.

This PR adds code to support an upper limit to the number of supported concurrent connections forwarded into the VM. By default there is no limit, but a command-line `--max-connections` allows it to be set to some concrete value. When the limit is hit, incoming connections are closed, matching the behaviour of `libuv` when it runs out of resources.

Note this PR only affects the code where connections are forwarded *into* the VM. Connections from the VM to the outside are unchanged: such outgoing connections are only limited by the maximum number of file descriptors in the `com.docker.slirp` process since they are multiplexed over a single virtual ethernet connection to the VM.

This PR also fixes a connection leak in some Mac error paths.

### DIFF
--- a/src/com.docker.slirp.exe/src/connect.ml
+++ b/src/com.docker.slirp.exe/src/connect.ml
@@ -5,20 +5,20 @@ let src =
 
 module Log = (val Logs.src_log src : Logs.LOG)
 
-let hvsockaddr = ref None
-
-let set_port_forward_addr x = hvsockaddr := Some x
-
-let max_connections = ref None
-
-let set_max_connections x = max_connections := x
-
 module Make(Time: V1_LWT.TIME)(Main: Lwt_hvsock.MAIN) = struct
   include Flow_lwt_hvsock_shutdown.Make(Time)(Main)
 
   open Lwt.Infix
 
   type address = unit
+
+  let hvsockaddr = ref None
+
+  let set_port_forward_addr x = hvsockaddr := Some x
+
+  let max_connections = ref None
+
+  let set_max_connections x = max_connections := x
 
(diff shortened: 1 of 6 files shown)


djs55:
@yomimono @samoht @avsm what do you think of this one?

In a separate commit I'd like to propose a limit of about 900-1000 for the Mac (where the virtio-vsock limit is 1024) and something similar for Windows, once I've figured out the AF_HYPERV limit.

samoht:
Seems that the `active_connections` and `max_active_connections` are in the `Make` functor for OSX but only `active_connection` is on Windows. Not sure that makes a big difference in the code we ship, but that means that 1/ the limit is set per functor instantiation (e.g you can instantiate 2 functors if you want to allocate more) and 2/ the max limit will be set only on the current functor on Windows.

Not sure all of that really matters thought :-)

djs55:
Yes that is a bit odd. I'll try to make it consistent at least (I've got a long term dream of merging both frontends together, so it'll be helpful if they look very similar)

djs55:
I'll merge this preemptively to try to get it into D4M/D4W beta 20 -- shout if you see any problems and I'll address them!

Issue #89, opened 2016-07-21
Simplify the build


djs55:
The build does work, but it's quite fiddly and is more CI-friendly than developer-friendly.

The build would be easier if we could:
- [x] combine `hostnet` and `proto-vmnet` since `proto-vmnet` is really a private protocol shared between vpnkit and hyperkit which we don't want to encourage anyone else to use. Ideally in the longer term we would switch away to something standard (PPP?) but in the meantime we should discourage any proliferation of this
- [x] ~somehow upstream `ofs` to datakit and install this from `opam`~ instead: made `ofs` into a sub-library within the project
- [x] remove `osx-daemon` and `osx-hyperkit` from this repo as they're no longer used here
- [x] somehow combine the Mac and Windows frontends: they have a lot of duplicated code which can diverge easily. The only problem is some Unix specifics currently in the Mac frontend

Then this project could become a "normal" project with just a Makefile and we could avoid using `opam` internally within the build. Developers would be free to use `opam` themselves as normal without worrying about clashing with the build runes in this repo.

djs55:
The build workflow for developers is now the same as other packages.

PR #90, opened 2016-07-21
Move Vmnet.ml{,i} into hostnet

djs55:
This small module isn't shared by anyone else and is really implementing
a private protocol shared between hyperkit and vpnkit. Ideally in the
future we would replace this with something standard (PPP?)

This patch simplifies the build a little by removing one of the dependencies
where the overhead of management (in terms of build and package metadata and
in terms of extra developer time taken to manage it) is greater than the
actual logic in the module.

Part of #89 


yomimono:
This seems sensible to me, and I welcome any change that reduces the complexity of our build.

djs55:
Thanks! Let's hope the CI is finally happy...

PR #91, opened 2016-07-21
Update to cstruct.2.2.0

djs55:
If we want to use [datakit](https://github.com/docker/datakit) in the future then we need to be in sync with its dependencies, including `cstruct.2.2.0`

The only real change was a [mirage-tcpip backport](https://github.com/djs55/mirage-tcpip/commit/ba91620972c565e4be2edfee41884ef7c6c3d8ee) but this cherry-picked relatively cleanly.


yomimono:
Looks great to me!

PR #93, opened 2016-07-22
Consolidate the 2 frontends (Mac and Windows)

djs55:
This PR is a bit big but this is mainly because it unifies things slowly, until at the end it deletes the Windows-specific frontend completely as the Mac frontend now works for both platforms.

This PR guarantees that both Windows and Mac versions support the same command-line arguments and roughly the same features -- previously the Windows frontend was beginning to diverge from the Mac one.

Most platform specific behaviour is triggered by command-line arguments e.g. on Windows we will request the use of Hyper-V sockets, while on Mac we will request the use of Unix domain sockets.

Note that `com.docker.slirp` frontend now does a little bit of branching depending on whether `Sys.os_type = "Win32"` for things like receiving privileged sockets on the Mac. Perhaps this could be moved into a command-line argument too?


src/com.docker.slirp/src/main.ml
```diff
@@ -44,72 +65,115 @@ let unix_listen path =
...
+      Unix.dup2 dev_null Unix.stderr;
+      Log.info (fun f -> f "Logging to Apple System Log")
+    end
   end;
   Log.info (fun f -> f "Setting handler to ignore all SIGPIPE signals");
-  Sys.set_signal Sys.sigpipe Sys.Signal_ignore;
-  set_nofile nofile;
-  Log.info (fun f -> f "vpnkit version %%VERSION%% with hostnet version %s %s and uwt version %s"
-    Depends.hostnet_version Depends.hostnet_pinned Depends.uwt_version
+  (try Sys.set_signal Sys.sigpipe Sys.Signal_ignore with Invalid_argument _ -> ());
```
yomimono: I'm curious about the circumstances under which this can fail and what the consequences are.

src/com.docker.slirp/src/main.ml
```diff
@@ -120,86 +184,131 @@ let main_t socket_path port_control_path max_connections vsock_path db_path nofi
...
-(* NOTE(aduermael): it seems to me that "/var/tmp/com.docker.slirp.socket" is a default value, right?
-This socket path is now dynamic, depending on current user's home directory. Could we just
-make it fail instead? In case no argument is supplied? *)
 let socket =
-  Arg.(value & opt string "/var/tmp/com.docker.slirp.socket" & info [ "socket" ] ~docv:"SOCKET")
+  let doc =
+    Arg.info ~doc:
+      "A URLs to connect to for ethernet of the form \
+      hyperv-connect://vmid/serviceid or hyperv-connect://vmid for the default serviceid on Windows \
+      or /var/tmp/com.docker.slirp.socket on Mac" ["ethernet"]
```
yomimono: Nit: this docstring is hard to parse; can we break it up into multiple sentences or rephrase it to avoid the substring "to connect to for"?

src/com.docker.slirp/src/main.ml
```diff
@@ -120,86 +184,131 @@ let main_t socket_path port_control_path max_connections vsock_path db_path nofi
...
-(* NOTE(aduermael): it seems to me that "/var/tmp/com.docker.port.socket" is a default value, right?
-This socket path is now dynamic, depending on current user's home directory. Could we just
-make it fail instead? In case no argument is supplied? *)
 let port_control_path =
-  Arg.(value & opt string "/var/tmp/com.docker.port.socket" & info [ "port-control" ] ~docv:"PORT")
+  let doc =
+    Arg.info ~doc:
+      "A URL to connect to for port control of the form \
+     hyperv-connect://vmid/serviceid on Windows or \
+     /var/tmp/com.docker.port.socket on Mac" ["port"]
```
yomimono: See above.

src/com.docker.slirp/src/main.ml
```diff
@@ -120,86 +184,131 @@ let main_t socket_path port_control_path max_connections vsock_path db_path nofi
...
+       to the VM." [ "vsock-path" ] ~docv:"VSOCK"
+  in
+  Arg.(value & opt string "" doc)
 
 let db_path =
-  Arg.(value & opt (some string) None & info [ "db" ] ~docv:"DB")
+  let doc =
+    Arg.info ~doc:
+      "A URLs to connect to datakitof the form \
+      file:///var/tmp/foo or tcp://host:port or \\\\\\\\.\\\\pipe\\\\irmin" ["db"]
```
yomimono: URLs -> URL, I think, and "datakitof" -> "datakit of", and could also use rephrasing as above.

yomimono:
Thanks very much for doing the unification work on this; it's much nicer not to have so much almost-duplicated code.  I stuck a few comments inline but generally I'm very happy to see this PR. :)

src/com.docker.slirp/src/main.ml
samoht: I think this could raise on Windows where the signal is not defined.

yomimono: Let me translate from Midwestern: there should be a comment here about why it's OK to ignore the exception.

djs55: A comment is a good idea. The interface is annoying. I think there is no such thing as SIGPIPE on Windows which is why the call fails, but it would be so much more convenient if when the program asks "please don't let me receive SIGPIPE signals" the Windows implementation could just go "yeah ok, that was easy" rather than raising!

src/com.docker.slirp/src/main.ml
djs55: I don't think what I've written makes any sense. I think I composed together words found in both the Mac and Windows forks, without checking the result for grammatical correctness.

PR #94, opened 2016-07-28
Avoid some scary errors

djs55:
Avoid logging an error when

- a `shutdown_write` encounters a `Uwt.Uwt_error` (previously we expected `Unix.Unix_error` but `uwt` uses different exceptions to `unix`)
- a `write` encounters an `EPIPE`: this means the remote has closed and we've only just found out


yomimono:
Good catches!  Looks fine to me.

avsm:
The lack of exhaustivity checking here is very annoying -- would it make sense to move the Uwt bindings to use Rresult combinators instead of raising exceptions?

PR #97, opened 2016-08-01
More DNS fixes

djs55:
- filter IPv6 DNS servers in the core library so it's easier to test; and to ensure covers both the `/etc/resolv.conf` path and the DNS-servers-in-database-keys path
- share the same primary/secondary server choosing logic between UDP and TCP: this fixes TCP DNS to the secondary IP by forwarding it to the second nameserver correctly

### DIFF
--- a/src/hostnet/lib/dns_forward.mli
+++ b/src/hostnet/lib/dns_forward.mli
@@ -6,3 +6,8 @@ module Make
     (Time: V1_LWT.TIME) : sig
   val input: secondary:bool -> ip:Ip.t -> udp:Udp.t -> src:Ipaddr.V4.t -> dst:Ipaddr.V4.t -> src_port:int -> Cstruct.t -> unit Lwt.t
 end
+
+val choose_server: secondary:bool -> (Ipaddr.t * int) list -> (string * (Ipaddr.t * int)) option
+(** [choose_server secondary servers] chooses an upstream server to use from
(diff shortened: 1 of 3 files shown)


src/hostnet/lib/dns_forward.mli
```diff
@@ -6,3 +6,12 @@ module Make
     (Time: V1_LWT.TIME) : sig
   val input: secondary:bool -> ip:Ip.t -> udp:Udp.t -> src:Ipaddr.V4.t -> dst:Ipaddr.V4.t -> src_port:int -> Cstruct.t -> unit Lwt.t
 end
+
+val only_ipv4: (Ipaddr.t * int) list -> (Ipaddr.t * int) list
+(** Since we don't currently support UDP to IPv6, filter all IPv6 nameservers
+    out of the list. *)
```
avsm: does this need to exposed in the interface? It doesn't appear to be used outside of Dns_forward

djs55: Good point, I was going to use it and then didn't. I'll remove it.

dsheets:
What about servers after the second? See for instance <https://forums.docker.com/t/docker-pull-timeout/15015/20>.

djs55:
Hm, perhaps we should expand the list of servers to 8 or 16?

djs55:
I filed #99

PR #101, opened 2016-08-18
Support more than 2 physical DNS servers

### LINKED ISSUE #99: Support more than 2 DNS servers
See this forum post: https://forums.docker.com/t/docker-pull-timeout/15015/20.

The poster has a list at least 4 elements long. We could expose perhaps 8 or 16 DNS IPs to the VM and map these onto the current number of external resolvers.

djs55:
This PR generalises the mechanism to include a list of server IPs rather than a single primary and single secondary.

The PR also adds another 8 virtual DNS IPs. Hopefully this will work better for people who have more than 2 real DNS servers -- this probably happens when connecting to a VPN, where you are likely to gain additional servers.

Fixes #99


djs55:
I'd better merge this since I'm about to refactor the tests and it would clash quite badly.

PR #104, opened 2016-08-22
test: temporarily disable most of the throughput tests

djs55:
Although the PR built and tested fine, the merged build keeps
timing out with no log output. Let's see if the throughput tests
are causing a problem.


djs55:
This patch was merged a while ago as part of another PR.

PR #105, opened 2016-08-25
Dynamically update search domains

djs55:
- read `search` domains from the database key `slirp/dns`
- dynamically change the DHCP server configuration to include this as option 119 (Note the requirement to DNS encode the name)
- reduce the DHCP lease time to 10s so the values are refreshed on a reasonable human timescale


yomimono:
A lease time of 10s means that clients will (by default) query every 5s; is that intended?

djs55:
Yeah -- the idea is to update it as soon as possible after the user hits "apply" in the Mac system preferences panel window. I guess the goal is to update it before the user thinks it hasn't worked, gets annoyed and files an issue :)

yomimono:
Yep, got that; it seems very noisy to me, but I can't find a nicer way to do this with DHCP. :/

PR #106, opened 2016-08-26
DNS: serve up entries from /etc/hosts

djs55:
This PR allows users to edit `/etc/hosts` (or the Windows equivalent, defender-permitting) and then resolve the names inside the VM as well as on the host.

The patches break down into:
- Parse `/etc/hosts` (or the Windows equivalent) and keep a cache of the values found therein. In case the path is not quite as universal as hoped, it can be overridden by a command-line argument `--hosts <path>`
- When a DNS request is received, first lookup `A` and `AAAA` requests in our cache. Reply with the any cache hit; otherwise fall back to upstream servers.
- A test for the DNS lookup through the cache
- A test for the `/etc/hosts` file parser

Note I tried using the `Fs_event` API from `libuv`/`uwt` but I couldn't get it to emit more than one event, so this PR falls back to polling every 5s.

### DIFF
--- a/src/com.docker.slirp/src/main.ml
+++ b/src/com.docker.slirp/src/main.ml
@@ -49,6 +49,7 @@ module Config = Active_config.Make(Host.Time)(Host.Sockets.Stream.Unix)
 module Forward_unix = Forward.Make(Connect_unix)(Bind)
 module Forward_hvsock = Forward.Make(Connect_hvsock)(Bind)
 module HV = Flow_lwt_hvsock.Make(Host.Time)(Host.Main)
+module Hosts = Hosts.Make(Host.Files)
 
 let file_descr_of_int (x: int) : Unix.file_descr =
   if Sys.os_type <> "Unix"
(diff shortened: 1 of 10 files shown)


samoht:
Did you look at https://github.com/samoht/irmin-watcher for file watching? This expose simple [hook](https://github.com/samoht/irmin-watcher/blob/master/src/irmin_watcher_core.mli#L11) that you can use to listen to a directory:

```
Irmin_watcher.hook 0 "/etc/host" (fun _ -> (* NEW CHANGE! *) )
```

If that's useful, I could probably rename the library to make it a bit less Irmin-centric ...

djs55:
That looks like it would probably work for the `Lwt_unix` backend. One of the problems we have using other libraries in `vpnkit` at the moment is the `uwt` backend -- it means we can't touch `Lwt_unix`-- all the threads block forever since there's no "engine" running; there's something else in `uwt` :( If the embryonic plan to use grand central dispatch (and IOCP?) works out then perhaps we could switch to that and it might be more interoperable.

djs55:
I manually tested this by editing the `C:\Windows\System32\drivers\etc\hosts` file and to my great surprise, it worked!

PR #107, opened 2016-08-30
Reduce the amount of DHCP log spam

djs55:
Now that we're using a short lease we generate a lot of logs: one
line per BOOTREQUEST and BOOTREPLY every half (?) lease interval.

This patch logs the first BOOTREQUEST and first BOOTREPLY as `info`
rather than `debug` since these are genuinely useful things to see
in the logs. Subsequent BOOTREQUEST and BOOTREPLY messages are
suppressed. Other message types are logged as `info`.


djs55:
I'd like to update D4M/D4W today to get more testing of the previous patches. I'll merge this to avoid complaints about log-explosions, but feel free to raise objections to this and I'll address them retrospectively.

PR #109, opened 2016-09-01
Reduce log spam

djs55:
Since the reducing of the DHCP lease time, the amount of log spam has increased. This PR tries to silence some of the worst offenders:
- remove a `printf` in mirage-tcpip (current upstream uses different logging now which we'll inherit when we move to it)
- stop printing arp debug messages by default (the arp implementation seems pretty solid)
- avoid double-processing the `BOOTREQUEST` messages

The latter is quite interesting: we currently hook the DHCP client in via a `Netif` listener, which allows it to send and receive raw frames. Indeed, the DHCP packet parsing function expects to see the whole ethernet frame. Unfortunately the TCPIP stack also receives the frame and complains that no-one is listening on port 67. When we migrate away from using the TCPIP stack, towards a more low-level driver of the TCP code we will be able to avoid double-processing these packets.

### DIFF
--- a/opam/darwin/packages/dev/tcpip.999/url
+++ b/opam/darwin/packages/dev/tcpip.999/url
@@ -1,1 +1,1 @@
-git: "git://github.com/djs55/mirage-tcpip#3.0.0-beta5"
+git: "git://github.com/djs55/mirage-tcpip#3.0.0-beta6"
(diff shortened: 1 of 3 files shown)


djs55:
Obsoleted by #155

PR #110, opened 2016-09-01
Remove unused packages osx-hyperkit, osx-daemon

djs55:
These are currently unused following the merge of the Windows frontend `com.docker.slirp.exe` with the Mac frontend `com.docker.slirp`.

The code should probably be moved to another repo/component. I'll look around and make a PR.

This is part of #89


djs55:
Hm there may be general unreliability in the windows networking tests, at least on the CI:
```
[hostnet: make test] Command started
+ make "test" (CWD=C:/projects/vpnkit/_build/opam/4.03.0+mingw64c/build/hostnet.local)
```
It might be something like a dialog popping up asking for permission to listen on a port? Or a genuine bug, can't tell :(

PR #111, opened 2016-09-05
Revert "dhcp: set the lease time to 10s"

djs55:
This reverts commit 6ea1b3809586748f0322651fa7704e7372d87256.

Unfortunately the short lease time has exposed a problem with the
mirage-tcpip stack, which we need to fix before we can do this.


djs55:
Since this is time-critical and it's a trivial revert, I'm going to merge this before the CI builds the PR. When the CI builds the master branch I'll update D4D.

PR #112, opened 2016-09-07
Flow handler fixes

MagnusS:
Update to djs55/mirage-tcpip, beta7 branch and fix flow handler to use correct src/dst. Also enable tcpip tests, as they should now work with the modified stack

mirage-tcpip beta 7 includes the following fixes:
- Fix on_flow_arrival handler bugs: djs55/mirage-tcpip#2
- Fix timer recovery bug: djs55/mirage-tcpip#1
- Update to mirage-vnetif 0.2.0
- Fix pcap recording race: djs55/mirage-tcpip#3

additional ci fix:
- Remove curl from installed cygwin tools and use the one installed in appveyor (otherwise curl does no longer seem to be in path...)


MagnusS:
Looks like this PR is finally green!

yomimono:
This looks *great* to me. :smile_cat:

PR #116, opened 2016-09-20
Increase UDP datagram limits

djs55:
The maximum UDPv4 datagram size is 65507 (which is IPv4 datagram (65535) - IPv4 header(20) - UDPv4 header(8)). Previously the code assumed the size was limited to the MTU of the ethernet network.

Furthermore, if we get a "partial" `recvfrom` (which means the packet was too big for the buffer) then log and continue. Previously we would shut down the listening thread which caused UDP forwarding to stop working.

Related to [docker/for-mac#674]

### DIFF
--- a/src/hostnet/_oasis
+++ b/src/hostnet/_oasis
@@ -14,7 +14,7 @@ Library hostnet
   Path: lib
   Pack: true
   Findlibname: hostnet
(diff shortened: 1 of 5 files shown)


djs55:
The `tcpip` tests randomly failed:
```
```

PR #117, opened 2016-09-23
Attempt to fix watermarking

djs55:
We substitute %%VERSION%% with the $(git rev-parse HEAD), and attempt
to substitute the %%*_PINNED%% properly.


djs55:
I'm really keen to fix the watermarking so we can see what version we're running...

PR #118, opened 2016-09-23
For DNS, make the UDP NAT rules one-shot

djs55:
Previously we would hold open a UDP NAT rule (and associated
socket) for 60s. This could waste a lot of sockets for DNS-heavy
workloads.

This patch closes the socket and removes the rule as soon as the
response is received. If the response is never received then the
GC will kick-in as before.

This is a short term patch before the next DNS rewrite.


djs55:
Another one of these:
```
```

PR #119, opened 2016-09-23
dhcp: remove the static map, allow exactly one dynamic IP

djs55:
Unfortunately switching the client from udhcpc to dhcpd causes a
problem because by default dhcpd will send a "duid" which charrua
doesn't recognise corresponds to the MAC address in the static
map. Since vpnkit currently only expects one client anyway,
switch to a dynamic range containing one element.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   xcode:
-    version: "7.3"
+    version: "8.0"
   environment:
     OPAMYES: "1"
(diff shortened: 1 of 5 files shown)


djs55:
```
```

djs55:
Uh-oh CircleCI says
```
Error: Your Xcode (7.3.1) is outdated
Please update to Xcode 8.0.
Xcode can be updated from the App Store.
```

PR #125, opened 2016-10-03
Capture both sides of DNS transactions

djs55:
Previously we would only see the VM half of UDP DNS transactions, since `module Capture` is placed inline on the virtual ethernet link. This PR injects datagrams corresponding to UDP DNS activity on the host directly into the `module Capture` so the debug traces will show the query and response from the upstream server too.

Now that the DNS capturing is working well, the default log level of the `Dns_forwarder` can be raised from `Debug` to `Info`.


djs55:
Rebased after all the recent churn, seems to be working well.

PR #127, opened 2016-10-09
Use upstream APIs

djs55:
Previously we used a modified version of the Mirage `Tcpip_stack_direct` which

- relied on the bug where TCP SYNs would be ACKed even if not addressed
  directly to us (e.g. if we're acting as an IP router)
- added a flow `Accept`/`Reject` API to avoid having to register
  65536 hashtable entries, one per port.

This PR changes the wiring as follows:

- incoming packets are switched based on destination IPv4 address to
  separate custom stacks
- each custom stack uses the regular interfaces, has a single IPv4
  address. We don't rely on the TCP SYN accept bug any more.
- each custom stack has a custom `Netif.listen` callback which
  pattern-matches on incoming packets and intercepts TCP SYNs.
  Before the SYN is forwarded, the stack will issue a `connect`.
- packets which don't match a switch rule are sent to a default
  callback which sets up a new stack and configures a switch port (c.f.
  the openvswitch control plane)

As a readability benefit, we now use pattern-matching everywhere rather than a mixture of `add_listener`-style callbacks and pattern-matching.

All the new state is visible over the 9P debug interface.

It should be possible to compile this code against more recent versions of the Mirage TCP/IP stack, but this PR doesn't attempt to do so.


djs55:
```
[ERROR] The sources of the following couldn't be obtained, aborting:
          - oasis.0.4.7
          - ocamlify.0.0.1
          - ocamlmod.0.0.8
          - ounit.2.0.0
        (This may be fixed by running 'opam update')
```

djs55:
All the D4M regression tests passed except the `nmap` test which printed
```
com.docker.slirp: [ERROR] SYN packet was not intercepted for port 2030
com.docker.slirp: [ERROR] 192.168.1.103:1094: failed to connect, sending RST: Socket.Tcp.connect tcp:192.168.1.103:1094: caught Uwt.Uwt_error(Uwt.ECONNREFUSED, "tcp_connect", "")
```
The actual `nmap` seems to have worked ok:
```
[STDOUT  ] 2016-10-10 10:00:19.136: Nmap scan report for 192.168.1.103
[STDOUT  ] 2016-10-10 10:00:19.137: Host is up (0.0079s latency).
[STDOUT  ] 2016-10-10 10:00:19.137: Not shown: 698 filtered ports, 298 closed po
rts
[STDOUT  ] 2016-10-10 10:00:19.137: PORT     STATE SERVICE
[STDOUT  ] 2016-10-10 10:00:19.137: 22/tcp   open  ssh
[STDOUT  ] 2016-10-10 10:00:19.137: 111/tcp  open  rpcbind
[STDOUT  ] 2016-10-10 10:00:19.137: 1023/tcp open  netvenuechat
[STDOUT  ] 2016-10-10 10:00:19.137: 2049/tcp open  nfs
[STDOUT  ] 2016-10-10 10:00:19.137: 
[STDOUT  ] 2016-10-10 10:00:19.138: Nmap done: 1 IP address (1 host up) scanned in 2.10 seconds
```
but the test case failed to ping:
```
[STDOUT  ] 2016-10-10 10:00:19.142: + ping -c 1 192.168.1.103
[STDOUT  ] 2016-10-10 10:00:19.145: PING 192.168.1.103 (192.168.1.103): 56 data bytes
[STDOUT  ] 2016-10-10 10:00:29.153: 
[STDOUT  ] 2016-10-10 10:00:29.154: --- 192.168.1.103 ping statistics ---
[STDOUT  ] 2016-10-10 10:00:29.154: 1 packets transmitted, 0 packets received, 100% packet loss
[STDOUT  ] 2016-10-10 10:00:29.154: + exit 1
```
-- probably we haven't got a pattern match for ICMP in the current code.

djs55:
Tests are looking ok now.

PR #128, opened 2016-10-10
Simplify the build

djs55:
Previously the regular build would need to use `opam` to build sub-packages
from this repo before finally building the binary. This patch switches back
to a single-project which means that we get incremental compilation and
a generally easier compile-test cycle (no more worrying about path-pinning
versus git-pinning)

Future patches will remove more of the upstream package divergence (or at
least make it optional) so we will eventually be able to publish this
package in opam.

Related to #89 


djs55:
Woo, build simplified!

PR #130, opened 2016-10-11
Use upstream APIs `allocate_frame` and `Udpv4.write`

djs55:
Previously we added APIs in a custom fork of mirage-tcpip
so we could customise the source address of the IP datagrams. We don't
need to do this now that we use multiple TCP/IP instances.

With this patch the code works against upstream tcpip.2.8.1


yomimono:
Excellent!

djs55:
Thanks, I'll merge this one and prepare some DNS-related PRs next!

PR #132, opened 2016-10-25
Use the new dns-forward forwarding library

djs55:
This PR removes the built-in DNS forwarding logic and replaces it with a use of the experimental `dns-forward` library.

User-observable differences include:

- we only need to supply one DNS IP to the VM (previously we supplied 8 and resolvers often ignored them)
- the DNS forwarding library supports the notion of "zones", commonly used in VPN configurations to lookup private names only in private nameservers
- requests are now multiplexed over 1 socket per upstream server by rewriting the request ids. Previously we multiplexed at the UDP level by using one source port per request. This reduces the number of sockets we need on the host.

Programmer-observable differences include:

- update to use `Result.result` rather than polymorphic variants
- add a small reimplementation of a subset of the as-yet-unreleased `Lwt_result` API. When the next `lwt` release happens we can remove this.

Still to-do:

- more testing, particularly of the zone support


yomimono:
This is awesome! I'm still looking through individual changes, but I'm really excited to see this PR.

src/hostnet/sig.ml
```diff
@@ -4,10 +4,12 @@ module type FLOW_CLIENT = sig
   type address
 
   val connect: ?read_buffer_size:int -> address
-    -> [ `Ok of flow | `Error of [ `Msg of string ] ] Lwt.t
+    -> flow Error.t
   (** [connect address] creates a connection to [address] and returns
       he connected flow. *)
 
+  val getclientname: flow -> address
+  (** Query the address the client is bound to *)
```
yomimono: What's the intended use of this?  It's stubbed out or generates runtime errors in many places, and `grep` didn't find any internal callers, so I wonder whether it's really needed.

src/hostnet/slirp.ml
```diff
@@ -335,45 +358,26 @@ module Make(Config: Active_config.S)(Vmnet: Sig.VMNET)(Resolv_conf: Sig.RESOLV_C
...
-        | x :: xs -> if compare x y = 0 then Some i else loop (i + 1) xs y in
-      loop 0
-
     (** Handle IPv4 datagrams by proxying them to a remote system *)
     let input_ipv4 t ipv4 = match ipv4 with
       (* Respond to ICMP *)
       | Ipv4 { raw; payload = Icmp _; _ } ->
         let none ~src:_ ~dst:_ _ = Lwt.return_unit in
-        Stack_ipv4.input t.endpoint.Endpoint.ipv4 ~tcp:none ~udp:none ~default:(fun ~proto:_ -> none) raw
+        let default ~proto:_ = Stack_icmpv4.input t.endpoint.Endpoint.icmpv4 in
```
yomimono: Should limit this to protocol 1 (ICMP).

src/hostnet/slirp.ml
```diff
@@ -431,10 +436,11 @@ module Make(Config: Active_config.S)(Vmnet: Sig.VMNET)(Resolv_conf: Sig.RESOLV_C
...
```
yomimono: Should limit this to protocol 1 (ICMP).

djs55:
Thanks for the quick response! Regarding `CONN` and `read_into`: do you suggest explicitly requiring `CONN` (== `FLOW` + `read_into`) in those (hopefully few) places where it's needed (all TCP-style forwarding loops) and then removing the dummied-out `read_into` from `Udp`? Sounds like a good move to me -- I'll investigate.

BTW here's the [socket-like interface](https://github.com/djs55/ocaml-dns-forward/blob/db0b032597da1dcabfd4c906609c2584da9a6799/lib/dns_forward.mli#L34) currently required by the DNS forwarder. It's really a small evolution of the interface we've got here. So far I've found it quite handy to treat UDP and TCP in the same way (similar to file descriptors on Unix with `connect` `read/recv` and `write/send`). In particular it made it easy to share the DNS request/response multiplexing code without having to have a different implementation for `connect/read/write` versus `sendto/recvfrom` IYSWIM

Regarding: `getclientname`: it's currently used by [dns-forward here](https://github.com/djs55/ocaml-dns-forward/blob/db0b032597da1dcabfd4c906609c2584da9a6799/lib/dns_forward_rpc.ml#L114) to find an address to call the [per-request response callback](https://github.com/djs55/ocaml-dns-forward/blob/db0b032597da1dcabfd4c906609c2584da9a6799/lib/dns_forward_rpc.ml#L38) with. The callback is used to make nice .pcap files in vpnkit. In practice this might not have worked as well as I'd hoped because the typical address is `0.0.0.0` which I'm having to special-case in the .pcap recorder anyway, so perhaps I should just delete this after all...

yomimono:
I dropped some more specific comments in as well.  It's probably worthwhile to take some time to review the overall set of module type signatures in `vpnkit` (and think about those in other OCaml libraries that handle abstracting over different connection types) -- I get the feeling we can reduce the apparent complexity of what's here currently.

I agree with you about the need for more testing. :)

Once tests are passing, I'm happy to merge this as I think it's a big improvement overall.

yomimono:
Sorry, didn't see your comment before leaving mine!

> BTW here's the socket-like interface currently required by the DNS forwarder. It's really a small evolution of the interface we've got here. So far I've found it quite handy to treat UDP and TCP in the same way (similar to file descriptors on Unix with connect read/recv and write/send). In particular it made it easy to share the DNS request/response multiplexing code without having to have a different implementation for connect/read/write versus sendto/recvfrom IYSWIM

Yes, that makes a lot of sense to me.

djs55:
Thanks again -- I'll do another round of updates and fix whatever's wrong with the CI.

I definitely agree w.r.t reviewing the module signatures used in this repo -- it feels like we haven't quite factored it in the ideal way yet :/

PR #133, opened 2016-10-28
Allow the DHCP server domain name to be overriden

djs55:
Previously the DHCP server hardcoded the domain name to `local`. This PR allows the domain name to be overriden via the database key `slirp/domain`.

Note this PR also contains a bug fix to the very similar search domain override mechanism. Ideally I'd add a unit test for this but it's a bit tricky using the built-in DHCP client in `tcpip.2.8.0` which doesn't seem to expose the information.

### DIFF
--- a/src/bin/main.ml
+++ b/src/bin/main.ml
@@ -235,6 +235,7 @@ let main_t socket_url port_control_url introspection_url max_connections vsock_p
       local_ip = Ipaddr.V4.of_string_exn "192.168.65.1";
       extra_dns_ip = [];
       get_domain_search = (fun () -> []);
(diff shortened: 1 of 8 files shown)


yomimono:
This looks good to me.

djs55:
Thanks!

djs55:
BTW I wonder if there's a nicer way to write this database-driven configuration stuff. Perhaps something reactive where we could take variables from the db and nicely combine them together with other bits of information to make a "variable" DHCP config rather than the slightly ad-hoc looking `Active_list` -> mutable reference -> `unit -> string` callback stuff I've got here.

PR #134, opened 2016-10-28
Include the Mac binary in the artefact, not a dangling symlink

djs55:
Previously the `mv main.native` simply moved the symlink to the binary
into the staging area. Using `cp main.native` means the symlink should
be followed and the contents copied instead.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -24,7 +24,7 @@ clean:
 
 stage:
 	mkdir -p _build/root/Contents/MacOS
-	mv main.native com.docker.slirp
+	cp main.native com.docker.slirp
 	mv com.docker.slirp _build/root/Contents/MacOS/com.docker.slirp
 	dylibbundler -od -b \
 	  -x _build/root/Contents/MacOS/com.docker.slirp \


djs55:
It looks like [this artefact](https://500-58395340-gh.circle-artifacts.com/0/Users/distiller/vpnkit/com.docker.slirp.tgz) is better:
```
$ tar -tvzf com.docker.slirp.tgz 
drwxr-xr-x  0 distiller staff       0 28 Oct 17:29 Contents/
drwxr-xr-x  0 distiller staff       0 28 Oct 17:29 Contents/MacOS/
drwxr-xr-x  0 distiller staff       0 28 Oct 17:29 Contents/Resources/
drwxr-xr-x  0 distiller staff       0 28 Oct 17:29 Contents/Resources/lib/
-rwxr-xr-x  0 distiller staff 11241544 28 Oct 17:29 Contents/MacOS/com.docker.slirp
```
-- that looks like sufficient bytes to contain a vpnkit binary rather than just a symlink!

djs55:
Thanks for the +1, I'll merge and get the CI building the artefact from master...

PR #135, opened 2016-11-01
Reduce spammy logging

djs55:
@samoht observed a recent test run generated 227MiB of log files, mostly "Errors" explaining that the remote server refused a connection (as it is totally at liberty to do)

This PR:
- increases minimum debug level to Info (was debug)
- downgrades some "errors" to "debug"


djs55:
I'd like to test this a bit manually with the stress test @samoht mentioned to check if the spam has been reduced enough.

djs55:
I ran the spammy test and it was much reduced: down to ~300 lines.

PR #136, opened 2016-11-01
Remove one buggy packet parser

djs55:
This PR removes the buggy `match.ml` in favour of the slightly-less-buggy `frame.ml`. It also makes the `frame.ml` code more robust. Any packets which are malformed are added to a virtual trace `bad.pcap` and may be read over the 9P diagnostic interface.

It would be nice to remove the remaining parser and replace it (and the variant type `Frame.t`) with something in `mirage-tcpip` or similar.

It would also be nice to transition more of the code to process `Cstruct.t list` values, since this would allow us to remove a `Cstruct.concat` in the `vmnet` code needed because the `CHANNEL.read_exactly` function returns a `Cstruct.t list` but the `V1_LWT` signatures have `type buffer = Cstruct.t`


djs55:
@yomimono here's the PR which fixes the `Cstruct` out-of-bounds exception seen in the stress test by removing the buggy code responsible for it :)

yomimono:
This patch looks good to me!

> It would be nice to remove the remaining parser and replace it (and the variant type Frame.t) with something in mirage-tcpip or similar.

Agreed! Also, we should think about moving the predicate-matching code for pcap recording into a separate package (or maybe into `pcap-format` itself?).  This is code that's generally useful and it'd be nice to have for other projects.

> It would also be nice to transition more of the code to process Cstruct.t list values

This is a great time to bring this up.  :)

PR #138, opened 2016-11-09
Allow a .tar to be downloaded containing the introspection fs

djs55:
This patch adds a command-line argument

  --diagnostics \\.\pipe\foobar
  --diagnostics /tmp/foobar

which will listen on a local Unix domain socket or Windows named pipe. Clients should connect to the socket or pipe and read until EOF. The data will be a tar-formatted diagnostics archive which contains the same data as the 9P introspection fs.


djs55:
This still needs a bit more testing.

djs55:
OK CI is green and this should be very useful for diagnostics, especially on Windows.

PR #139, opened 2016-11-10
test: check that UDP source ports are preserved

djs55:
This test is based on the example from [docker/for-mac#926] where we have 2 external echo servers and one internal client/server socket. The internal socket sends to both echo servers and records the source ports of the replies.

This triggers a bug where the initial order of transmission wins, since the source port is captured at the time the forwarding socket and shared NAT rule is created.

I don't expect this test to pass currently.

### DIFF
--- a/src/hostnet_test/nat.ml
+++ b/src/hostnet_test/nat.ml
@@ -63,28 +63,39 @@ module Make(Host: Sig.HOST) = struct
   end
 
   module UdpServer = struct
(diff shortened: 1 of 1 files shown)


djs55:
Hm, suspicious that the build passed. Will investigate.

djs55:
With tests re-enabled in the CI, this test fails as expected.

djs55:
Merged as part of #147

PR #140, opened 2016-11-10
Merge the two Makefiles together

djs55:
Previously we had a confusing Makefile and src/Makefile, one
for the CI and one for development. This patch attempts to merge
both together so `make` in the root should "just work" (TM)

This patch also splits the `opam` file in the repo which contains
a set of pinned versions, from a more general `src/opam` which
is suitable for `opam pin`. Eventually these two will merge also.


djs55:
Since this is just a build change, I'll merge it.

PR #143, opened 2016-11-11
Use the protocol family which matches the address

djs55:
Previously we used `Ipaddr.t` values everywhere which allowed both v4 and v6, but we always used functions like `Uwt.Tcp.init ()` which only works for v4 leading to an `EINVAL` on `connect`. This PR chooses the correct protocol family for the address.

This PR also corrects log lines which were previously hardcoded to say things like `Socket.TCP4`.

### DIFF
--- a/src/hostnet/host_lwt_unix.ml
+++ b/src/hostnet/host_lwt_unix.ml
@@ -82,8 +82,8 @@ let sockaddr_of_address (dst, dst_port) =
 
 let unix_bind_one pf ty ip port =
   let protocol = match pf, ty with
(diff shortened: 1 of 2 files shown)


djs55:
Building the compiler failed on CircleCI:
```
        # Error: Files ../parsing/parse.cmx and ../parsing/parser.cmx
        #        make inconsistent assumptions over implementation Parser
```
I'll trigger another build by force-pushing a rebased branch.

djs55:
Unfortunately I don't know how to test this as I don't have an IPv6 network. It's definitely necessary but may not yet be sufficient for IPv6 DNS support...

PR #144, opened 2016-11-11
Update to dns-forward.0.5.1

djs55:
This fixes a bug where cached responses (and those from the local
names callback) accidentally were tagged as DNS requests rather
than responses. It's amazing how much software is robust to this.


djs55:
This update has made building docker much more reliable under Docker for Mac.

PR #146, opened 2016-11-11
CircleCI: attempt to fix artifact upload

djs55:
Unfortunately it's not a fatal error for an artifact to be missing.

Previously we placed the .tgz file in .. believing it to be relative
to the -C src directory provided to tar. This patch places the .tgz
in the root of the tree.


djs55:
Artifact is back!

PR #151, opened 2016-11-14
test: add explicit 10s timeouts for most of the tests

djs55:
Previously the tests could block forever, particularly infuriating as
this locks the appveyor CI slot for 1hr. All the tests should execute
within 10s, except the max connections one which has been given a 4
minute maximum.


djs55:
10s was too short for the CI machines -- let's try 60s.

djs55:
OK, both CIs seem happy with this now.

PR #152, opened 2016-11-15
Update dns-forward to v0.7.0

djs55:
- Fix a (rare) race which could cause the wrong response to be
  used after a timeout
- Minimise transaction id reuse


djs55:
The "10 forwarded connections" test seems a bit intermittent. For context this has nothing to do with the TCP/IP stack, it actually only concerns the 9P protocol for opening up port forwards. I suspect there's a file descriptor handling bug somewhere in the harness.

At least we have some logs:
```
-- Forwarding with Lwt.001 Failed --
Check speed of 10 forwarded connections.
_build/_tests\Forwarding with Lwt.001.output:
main_lwt.native: [INFO] attempting a best-effort bind of ::1:1234
main_lwt.native: [INFO] attempting a best-effort bind of ::1:1234
main_lwt.native: [DEBUG] 127.0.0.1:1234: connecting
Tcp.Segment: TCP retransmission on timer seq = 452873015
main_lwt.native: [ERROR] PPP.listen: caught unexpected (Failure "Vmnet connection is disconnected"): disconnecting
Tcp.Segment: TCP retransmission on timer seq = 452873015
Tcp.Segment: TCP retransmission on timer seq = 452873015
Tcp.Segment: TCP retransmission on timer seq = 452873015
main_lwt.native: [DEBUG] Hostnet_udp udp:192.168.65.2:40583-8.8.8.8:53: expiring UDP NAT rule
main_lwt.native: [ERROR] Hostnet_udp udp:192.168.65.2:40583-8.8.8.8:53: caught unexpected exception Unix.Unix_error(Unix.EBADF, "check_descriptor", "")
Tcp.Segment: TCP retransmission on timer seq = 452873015
Tcp.Segment: Max retransmits reached for connection - terminating
main_lwt.native: [ERROR] Socket.TCPV4.read 127.0.0.1:1234: caught Lwt.Canceled returning Eof
main_lwt.native: [ERROR] Socket.Stream: caught Unix.Unix_error(Unix.EBADF, "check_descriptor", "")
main_lwt.native: [ERROR] Socket.Stream: caught Unix.Unix_error(Unix.EBADF, "check_descriptor", "")
[failure] timeout
[failure] timeout
The full test results are available in `_build/_tests`.
```
I'm a bit suspicious of the Tcp stack warnings there because this test shouldn't involve the stack at all. This may be state left over from a previous test.

Since this PR is urgent, I'll merge despite this failure.

PR #154, opened 2016-11-16
README.md: add brief notes on using with hyperkit

djs55:
This describes the minimal command-line arguments needed to start
`com.docker.slirp` and to connect `com.docker.hyperkit` to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,17 @@ Build the application using:
 make
 ```
 
+Running with hyperkit
+---------------------
+
+First ask `com.docker.slirp` to listen for ethernet connections on a local Unix domain socket:
+```
+com.docker.slirp --ethernet /tmp/ethernet --debug
+```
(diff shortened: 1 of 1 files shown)


djs55:
@justincormack is this the kind of thing you were looking for?

justincormack:
Yeah that looks great, thanks.

ijc25:
LGTM.

In my (old) notes about this I say it can be just `com.docker.slirp --debug` and `-s 2:0,virtio-vpnkit`, suggesting that the default socket path was the same when I wrote those notes (mtime is 2 June 2016).

Looks like hyperkit uses `/var/tmp/com.docker.slirp.socket` as the default, but vpnkit seems to not (anymore?) and I can't tell what if any default it has, it mentions `/var/tmp/com.docker.slirp.socket` in the help .

PR #155, opened 2016-11-17
Trim or demote spammy errors

### LINKED ISSUE #108: Remove unnecessary log spam
Since the DHCP lease time was reduced we're still seeing quite a lot of regular spam e.g.

```
com.docker.slirp: [DEBUG] UDP 0.0.0.0:68 -> 255.255.255.255:67 len 308
com.docker.slirp: [DEBUG] Socket.Datagram.input 0.0.0.0:68-255.255.255.255:67: ignoring broadcast packet
ICMP Destination Unreachable: Destination port unreachable
```

djs55:
The goal is to remove as many "expected" logs which are emitted with severity "error", so it becomes possible to grep the logs for "ERROR" and see something meaningful.

Fixes #108

### DIFF
--- a/src/bin/main.ml
+++ b/src/bin/main.ml
@@ -105,10 +105,12 @@ let hvsock_connect_forever url sockaddr callback =
   aux ()
 
 let start_introspection introspection_url root =
-  Lwt.async (fun () ->
-    log_exception_continue "start_introspection_server"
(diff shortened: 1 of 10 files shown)


djs55:
Intermittent CI failure: #156

PR #164, opened 2016-11-30
Use the correct Unix.inet_addr matching the PF in bind

djs55:
The Unix.inet_addr type includes both IPv4 and IPv6. When we send
traffic over UDP we first bind the socket to "any". Previously we
used the value `Unix.inet_addr_any` but this is IPv4-only. This patch
selects either `Unix.inet_addr_any` or `Unix.inet6_addr_any`.


djs55:
@mor1 reported this fixed v6 DNS for him.

PR #168, opened 2017-01-05
Disable -warn-error by default

djs55:
Ideally we'd like to enable -warn-error for developer builds but disable
-warn-error for tagged release builds. This patch disables -warn-error
for both cases: it's not worth too much work on the current build and
packaging system when we're likely to switch over to topkg soon.

This allows vpnkit to build with lwt.2.7.0.

Related to [ocaml/opam-repository#8195]
Related to [ocaml/opam-repository#8196]


dsheets:
Perhaps you could just remove warning 3 (deprecated feature) from the warning-as-error list? Most others are safe, I believe, and help keep the codebase clean by making warnings unmissable.

djs55:
@dsheets good idea, I'll do that.

PR #188, opened 2017-02-22
Fix ethernet parsing bug, add l2 bridge

MagnusS:
Adds an optional bridge with mirage-vnetif. Also fixes a bug where ethernet frames were read with the `src`/`dst` fields swapped and adds a `.merlin` file to the project (this PR subsumes #187).

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -1,0 +1,15 @@
+PKG protocol-9p protocol-9p.unix cstruct cstruct.lwt lwt.unix io-page.unix logs
+PKG mirage-types.lwt astring named-pipe.lwt
+PKG cmdliner alcotest lwt.unix logs.fmt dns.mirage lwt.preemptive
+PKG hvsock hvsock.lwt-unix datakit-server.fs9p win-eventlog asl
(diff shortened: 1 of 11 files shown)


.merlin
```diff
@@ -4,6 +4,7 @@ PKG cmdliner alcotest lwt.unix logs.fmt dns.mirage lwt.preemptive
 PKG hvsock hvsock.lwt-unix datakit-server.fs9p win-eventlog asl
 PKG ipaddr mirage-flow ppx_sexp_conv pcap-format mirage-console.unix tcpip
 PKG charrua-core dns uwt lwt threads dns-forward tar
+PKG mirage-vnetif
```
djs55: Do you need to run `make commit` in the `repo` directly to add the `mirage-vnetif` package to the opam universe (as used by CI)?

MagnusS: @djs55 it looks like it built without it. Should I run `make commit` anyway to mirror the package in the local repo?

MagnusS: Looks like it was already there due to the tcpip-tests

PR #189, opened 2017-02-28
Add bridge parameter

MagnusS:
Adds a `--default-bridge` parameter to set bridge mode when no db is available.


MagnusS:
Looks like the circleci build is green and github shows the wrong status

PR #191, opened 2017-03-02
Capture up to 64KiB of NTP traffic for debugging

djs55:
This captures UDP packets whose source or destination port is 123.


djs55:
CircleCI succeeded but the event must have gotten missing.

PR #195, opened 2017-03-10
Update dns-forward to 0.8.2

djs55:


djs55:
I think this version of dns-forward might have a defect: it seems willing to declare a server offline too readily. I'm investigating.

PR #196, opened 2017-03-10
If no servers are configured, use `getaddrinfo` 

djs55:
If no upstream servers are configured we switch to using the `getaddrinfo` API. This is probably fine for `A` and `AAAA` records but won't work for more exotic queries. However more exotic queries are probably better directed towards specific known-good servers rather than default local ones.


djs55:
There's one user report that this strategy has fixed the DNS problem: https://github.com/docker/for-mac/issues/1397#issuecomment-285919564

PR #200, opened 2017-03-17
Add a command-line argument for the db configuration branch

djs55:
Previously we hard-coded the branch name but it's useful to be able to choose via the command-line.


djs55:
This has been working pretty well in my testing.

PR #207, opened 2017-04-06
Reuse old connections with same uuid

MagnusS:
Adds support for the previously ignored uuid field so clients can reconnect with the same mac/ip if disconnected and can suggest a preferred ip. Also adds bounds checking for the assigned IPs.

Updates dependencies to add `uuidm`


src/hostnet/slirp.ml
```diff
@@ -738,11 +744,15 @@ module Make(Config: Active_config.S)(Vmnet: Sig.VMNET)(Dns_policy: Sig.DNS_POLIC
...
          match parse [ buf ] with
-         | Ok (Ethernet { src = eth_src ; dst = eth_dst ; _ }) when
+         | Ok (Ethernet { src = eth_src ; dst = eth_dst ; payload }) when
             (not (Macaddr.compare eth_dst client_macaddr = 0 ||
                   Macaddr.compare eth_dst server_macaddr = 0 ||
                   Macaddr.compare eth_dst Macaddr.broadcast = 0)) -> (* not to server, client or broadcast.. *)
            if use_bridge then begin
+               (match payload with
+               | Ipv4 { src = ip_src ; _ } ->
+                       Log.debug(fun f -> f "hyrrah");
```
djs55: Is this a debug hunk accidentally included?

src/hostnet/slirp.ml
```diff
@@ -1058,6 +1060,25 @@ module Make(Config: Active_config.S)(Vmnet: Sig.VMNET)(Dns_policy: Sig.DNS_POLIC
...
+
+            (* check if a specific IP is requested *)
+            let preferred_ip = 
+                let uuid_bytes = Uuidm.to_bytes uuid in
+                let uuid_prefix = Bytes.sub uuid_bytes 0 (Bytes.length default_uuid_preferred_ip_prefix) in
+                if (Bytes.compare uuid_prefix default_uuid_preferred_ip_prefix) = 0 then
+                begin
+                    let uuid_suffix = Bytes.sub uuid_bytes 12 4 in
+                    let preferred_ip = Ipaddr.V4.of_bytes_exn uuid_suffix in
+                    Log.info (fun f -> f "Client requested IP %s" (Ipaddr.V4.to_string preferred_ip));
```
djs55: Rather than encode the preferred IP within the UUID, perhaps it would be simpler if we extended the vmnet protocol a little -- we could add a new command type e.g.
```
module Command : sig

  type t =
    | Ethernet of string (* 36 bytes *)
    | Bind_ipv4 of Ipaddr.V4.t * int * bool
    | Preferred_ipv4 of string (* 36 byte uuid *) * Ipaddr.V4.t (** <-- NEW *)
```
Separately we could have a hyperkit patch to add a new key=value pair to the `virtio-vpnkit` driver, so we could have command-lines like this:
```
-s 2:0,virtio-vpnkit,uuid=...,path=...,macfile=....,preferred-ip=...
```
which would then send the new message.

src/hostnet/slirp.ml
MagnusS: Yes - will remove

src/hostnet/slirp.ml
MagnusS: Yes, I agree this would be a better approach. As discussed directly I've added an issue tracking what needs to be done to have a separate `preferred_ip` command in https://github.com/docker/vpnkit/issues/209

MagnusS:
As discussed with @djs55 directly I'm merging this PR for now. The remaining changes to include the `preferred_ip` command in the vmnet protocol is tracked in #209

PR #210, opened 2017-04-11
Allow client to not send UUID

MagnusS:
If the client sends \0 instead of UUID, generate UUID on behalf of
client.


avsm:
lgtm

PR #211, opened 2017-04-19
DNS: don't attempt to look up localhost.local

djs55:
When performing a `docker push` to `localhost`, `docker` will look up both `localhost` and `localhost.local`. Unfortunately the Mac will seriously attempt to look up `localhost.local` using mDNS which takes several seconds to timeout.

This patch special-cases `localhost.local`, making it always return a failure. This allows lookups of other .local names to continue working.

Related to docker/for-mac#1536


djs55:
I pushed a rebase to trigger the CIs.

PR #216, opened 2017-04-26
Avoid reversing and duplicating DNS RRs

djs55:
When using the host resolver on the Mac, if we queried a DNS name `a` where:
```
a IN CNAME b
b IN CNAME c
c IN CNAME d
d IN A p.q.r.s
```
Then the binding to the `DNSServiceRef` API would return `[c; b; a]` (Note it seems to return only `CNAME`s in this situation). The new version of the binding `dnssd.0.3` preserves the order and returns `[a; b; c]`.

The function which would look up the terminal records (e.g. `A` or `AAAA`) would be called recursively on each `CNAME` which would cause `CNAME` and `A` records to be duplicated. This PR avoids adding duplicate entries to the list.

This is related to [docker/for-mac#1569]


djs55:
I think this slightly strange logic to recursively call `DNSServiceRef` should probably be moved to the `dnssd` package itself. I'll do this in a later PR.

PR #218, opened 2017-05-02
Update the documentation

djs55:
- add a Windows and Mac example plumbing diagram
- add `SOCK_STREAM` and `SOCK_DGRAM` to the main `vpnkit` diagram
- link to hyperkit and Mirage TCP/IP
- add a high-level walk-through describing what happens when a process in a Linux VM connected to `vpnkit` attempts to establish a TCP connection

### DIFF
--- a/docs/ethernet.md
+++ b/docs/ethernet.md
@@ -15,12 +15,24 @@ interface inside the VM. The virtual hardware implementation connects to
 `vpnkit` on the host over a Unix domain socket and encapsulates frames using a
 simple custom protocol.
 
+![Mac plumbing diagram](http://moby.github.io/vpnkit/mac.png)
+
 ## Inside Docker for Windows
 
 The Docker for Windows VM is running on top of Hyper-V. `vpnkit` on the host
 uses Hyper-V sockets to connect to a process inside the VM which accepts the
 connection and configures a `tap` device. Frames are encapsulated using the
 same custom protocol as on the Mac.
+
+![Windows plumbing diagram](http://moby.github.io/vpnkit/win.png)
+
+Note: the connection is currently made from the Host to the VM to work around
+a bug in older versions of Windows 10. At some point this should change to be
+a connection from the VM to the host.
+
+Note: the userspace process in the VM which configures a `tap` device could be
+replaced with a custom Linux kernel driver which knows how to encapsulate the
+frames and communicate over Hyper-V sockets.
 
 ## When a frame arrives in vpnkit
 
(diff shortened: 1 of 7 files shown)


docs/ethernet.md
```diff
@@ -1,28 +1,118 @@
...
-- if they contain IPv4 datagrams then we create a fresh TCP/IP stack, a fresh
-  switch port and connect them together so that all future IPv4 traffic to the
-  same destination address is processed by the new stack.
+  table
+- if they contain IPv4 datagrams then we create a fresh virtual TCP/IP endpoint
+  using the [Mirage](https://mirage.io/) [TCP/IP stack](https://github.com/mirage/mirage-tcpip)
+  (no kernel TCP/IP interfaces are involved), a fresh
+  switch port on our internal switch and connect them together so that all future IPv4 traffic to the
+  same destination address is processed by the new endpoint.
+
```
MagnusS: There's also the L2 bridge in there now - so frames can come from any other Vpnkit instance via the bridge or from the connected VM: https://github.com/moby/vpnkit/blob/master/src/hostnet/slirp.ml#L731
 -- and frames from the VM without server/client/broadcast dst addresses will be sent to the bridge https://github.com/moby/vpnkit/blob/master/src/hostnet/slirp.ml#L749

The static ARP table also includes mac addresses to other Vpnkit instances.

djs55: Cool -- are you volunteering to write a little bit about it? :) Perhaps I'll merge this and then you could make a PR on top?

MagnusS: Sure, I can add more details after this is merged

docs/ethernet.md
```diff
@@ -1,28 +1,118 @@
...
+Note that from the host kernel's point of view, there is no network connection
+to the VM and no set of associated firewall rules or routing tables. All outgoing
+connections originate from the `vpnkit` process. If the user installs some
+advanced networking or VPN software which reconfigures the routing table or
+firewall rules, it will not break the connection between `vpnkit` and the VM.
 
-The active ports may be queried over the 9P debug interface via the `ports`
-subdirectory.
+This technique for forwarding network connections is commonly known as
+[Slirp](https://en.wikipedia.org/wiki/Slirp).
```
samoht: The wikipedia page says:
>  Even though the Slirp-derived code was heavily criticized,[11] to the date there is no competing implementation available.

Maybe someone can try to update that page? :p /cc @amirmc

PR #224, opened 2017-05-17
AUTO: Update upstream packages

djs55:


samoht:
Do you know why the license file is needed? It seems to be available upstream: https://github.com/janestreet/jbuilder/blob/master/LICENSE.txt

djs55:
IIRC the script failed to find the archive. I ought to debug it, but I wanted a build asap...

samoht:
no pb, I was just curious :-) maybe because it's a txt file instead of md

PR #230, opened 2017-05-31
Improve logging and make the tests easier to debug

djs55:
- remove unnecessary error logs (triggered by the tests)
- only test the `Uwt` backend since the `Lwt_unix` one is known broken
- run the tests with `List.iter` so we can associate the log output with the test that generated it

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,9 +50,7 @@ vpnkit.exe: src/bin/depends.ml
 
 .PHONY: test
 test:
-	jbuilder build src/hostnet_test/main_lwt.exe
 	jbuilder build src/hostnet_test/main_uwt.exe
-	./_build/default/src/hostnet_test/main_lwt.exe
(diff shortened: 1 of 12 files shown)


djs55:
It fails on appveyor with
```

**** Starting test Ports: 1 port forward
Starting test case Perform an HTTP GET through a port forward
Fatal error: exception Unix.Unix_error(Unix.EADDRINUSE, "bind", "")
Raised at file "src/core/lwt.ml", line 805, characters 22-23
Called from file "uwt.ml", line 1835, characters 10-23
Called from file "list.ml", line 77, characters 12-15
Called from file "list.ml", line 77, characters 12-15
Called from file "src/hostnet_test/main_uwt.ml", line 19, characters 2-249
WARNING: Io_page on Windows doesn't guarantee alignment
SOMAXCONN = 2147483647
make: *** [Makefile:54: test] Error 2
```

PR #236, opened 2017-06-08
Add vpnkit-forwarder / vpnkit-expose-port

MagnusS:
Imports history of `vpnkit-forwarder` (was: proxy-vsockd)  and `vpnkit-expose-port` (was: slirp-proxy) from LinuxKit.

`vpnkit-forwarder` connects to the 9p port mount in the guest and forwards traffic to/from vpnkit over vsock to the host.

`vpnkit-expose-port` is a userland proxy used by `dockerd` to set up the port forwarding on demand via the 9p port mount.

Changes:
- renamed the binaries to `vpnkit-forwarder` and `vpnkit-expose-port`
- created `go/cmd` for the cli tools and `go/pkg/libproxy` for the library
- added a new Makefile and container build (`make build-in-container`)
- removed some of the unused/outdated files from the history, like the original Makefile, Dockerfile, README and LICENSE.


djs55:
Good to see both ends of the 9P protocol back together at last!

LGTM

PR #241, opened 2017-06-13
Attempt to fix CircleCI

djs55:
The `brew install` step is failing because:

>  Error: libffi 3.0.13 is already installed
>  To upgrade to 3.2.1, run `brew upgrade libffi`


djs55:
Merging without waiting for appveyor since this is a CircleCI-only fix.

PR #244, opened 2017-06-22
Add go/hooks/post_push to push an custom tag

djs55:
Docker Cloud autobuilds support hook scripts which can be used to push custom tags. The LinuxKit convention is to tag binary images with the git tree hash of the source code i.e. the hash of the contents and not the history (the commit). This means that if more commits happen but the contents don't change, the tag remains the same too.


djs55:
This might take a very experiments to get right. Let's see.

PR #245, opened 2017-06-22
Fix the linux image post_push script

djs55:
Previously the script failed to compute the tree SHA because the `git ls-tree` in the Docker Cloud environment didn't like the path `../go`. It works fine as `go` if we `cd ..` first.

### DIFF
--- a/go/hooks/post_push
+++ b/go/hooks/post_push
@@ -2,8 +2,12 @@
 
 echo hooks/post_push: attempting to compute the tree SHA of the current directory
 pwd
+CURDIR=$(pwd)
+# The git ls-tree HEAD ../<foo> fails on Docker Cloud's hook environment.
+# It works fine from the parent directory.
+cd ..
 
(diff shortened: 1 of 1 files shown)


djs55:
For future reference @ijc wrote in [Docker community slack](https://dockercommunity.slack.com/archives/C51K0GDU7/p1498136998712901):

> In a Jessie container:
```root@d8a8772e04ad:/lk/pkg/containerd# git --version
git version 2.1.4
root@d8a8772e04ad:/lk/pkg/containerd# git ls-tree HEAD -- ../containerd && echo $?
0
```
> But in a Stretch one:
```root@899423f7a9ed:/lk/pkg/containerd# git --version
git version 2.11.0
root@899423f7a9ed:/lk/pkg/containerd# git ls-tree HEAD -- ../containerd && echo $?
040000 tree b50181bc6e0084e5fcd6b6ad3cf433c4f66cae5a    ./
0
```


For reference this CI has `git version 1.9.1` ... older than Jessie

PR #246, opened 2017-06-23
travis: update to docker 17.05 for multistage build

MagnusS:
Try to upgrade Docker to 17.05 to see if we can use the multi-stage build instead of maintaining two sets of Dockerfiles.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,4 @@ before_install: # update to 17.05 to get multistage builds
   - sudo apt-get -y install docker-ce=17.05.0~ce-0~ubuntu-trusty
 
 script:
-  - cd go && make multi-stage
+  - cd go && make build-in-container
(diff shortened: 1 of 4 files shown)


MagnusS:
Looks like it works! Should I remove the other Dockerfiles in this PR as well?

djs55:
Let's simplify and delete as much as possible-- do we need the other Dockerfiles? I guess we already have `Makefile` targets for building the binaries if you want to do that separately.

MagnusS:
I don't think so, I guess Docker cloud is using the multistage build as well - so removing

MagnusS:
All green! But looks like I've lost merge access here

djs55:
I’ve sent you an invite link. If you accept it I think you should get merge access.

MagnusS:
@djs55 thanks!

PR #248, opened 2017-06-28
go: split into 2 images

djs55:
- `Dockerfile`: builds `vpnkit-forwarder`
- `Dockerfile.expose-port`: builds `vpnkit-expose-port`

This will allow one to become a linuxkit service and the other to be part of system init.

### DIFF
--- a/go/Dockerfile.expose-port
+++ b/go/Dockerfile.expose-port
@@ -5,7 +5,7 @@ RUN apk add --no-cache go musl-dev build-base
 ADD . /go/src/github.com/moby/vpnkit/go
 WORKDIR /go/src/github.com/moby/vpnkit/go
 
-RUN GOPATH=/go make build/vpnkit-forwarder.linux build/vpnkit-expose-port.linux
+RUN GOPATH=/go make build/vpnkit-expose-port.linux
 
 FROM scratch
-COPY --from=mirror /go/src/github.com/moby/vpnkit/go/build/vpnkit-expose-port.linux /vpnkit-expose-port
+COPY --from=mirror /go/src/github.com/moby/vpnkit/go/build/vpnkit-expose-port.linux /usr/local/bin/vpnkit-expose-port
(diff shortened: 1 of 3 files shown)


djs55:
@magnus: does this look like you'd expect? I've set up another cloud autobuild from this repo but referencing the `Dockerfile.expose-port`. I'm curious to see whether both images will be built.

go/Dockerfile.expose-port
```diff
@@ -0,0 +1,11 @@
...
+
+RUN apk add --no-cache go musl-dev build-base
+
+ADD . /go/src/github.com/moby/vpnkit/go
+WORKDIR /go/src/github.com/moby/vpnkit/go
+
+RUN GOPATH=/go make build/vpnkit-forwarder.linux build/vpnkit-expose-port.linux
+
+FROM scratch
+COPY --from=mirror /go/src/github.com/moby/vpnkit/go/build/vpnkit-expose-port.linux /vpnkit-expose-port
```
MagnusS: Since we're going to use `vpnkit-expose-port` as an init-container in linuxkit, it's probably better to put it in a location that makes sense when extracted in the host fs. E.g. `/usr/local/bin/vpnkit-expose-port`.

go/Dockerfile.expose-port
```diff
@@ -0,0 +1,11 @@
+FROM linuxkit/alpine:630ee558e4869672fae230c78364e367b8ea67a9 AS mirror
+
+RUN apk add --no-cache go musl-dev build-base
+
+ADD . /go/src/github.com/moby/vpnkit/go
+WORKDIR /go/src/github.com/moby/vpnkit/go
+
+RUN GOPATH=/go make build/vpnkit-forwarder.linux build/vpnkit-expose-port.linux
```
MagnusS: I think we only need the build/vpnkit-expose-port.linux target here?

MagnusS:
@djs55 just some minor comments, lgtm :-)

Do you think the iptables-wrapper should go in the expose-port container as well?

djs55:
@MagnusS Thanks! I've made the fixes you suggest. Adding the `iptables-wrapper` to `vpnkit-expose-port` sounds fine -- they're both part of the same "port plugin" for Docker.

djs55:
Unfortunately the travis image is broken with
```
E: Failed to fetch https://debian.neo4j.org/repo/stable/Packages.gz  Hash Sum mismatch
```
-- this seems to be a separate issue.

djs55:
I don't think we need to wait for appveyor since this code isn't built on Windows.

PR #250, opened 2017-07-03
Don't disable DNS queries for .local

djs55:
Previously we disabled queries for `*.local` to avoid using (slow) mDNS but some users do use this domain and need it to work.

This PR
- removes the workaround so `*.local` queries will work
- sets the default domain name to `localdomain` rather than `local` so we don't trigger mDNS by accident

Related to [docker/for-mac#1811]

### DIFF
--- a/src/bin/main.ml
+++ b/src/bin/main.ml
@@ -279,7 +279,7 @@ let main_t socket_url port_control_url introspection_url diagnostics_url max_con
       local_ip;
       extra_dns_ip = [];
       get_domain_search = (fun () -> []);
-      get_domain_name = (fun () -> "local");
+      get_domain_name = (fun () -> "localdomain");
       global_arp_table;
       client_uuids;
       bridge_connections = true;
--- a/src/hostnet/host_lwt_unix.ml
+++ b/src/hostnet/host_lwt_unix.ml
@@ -810,16 +810,13 @@ module Dns = struct
       | _ -> acc
     ) [] x
 
-  let is_mdns name =
-    match List.rev @@ Dns.Name.to_string_list name with
-    | [] -> false
-    | n :: _ -> n = "local"
+  let localhost_local = Dns.Name.of_string "localhost.local"
 
   let resolve question =
     let open Dns.Packet in
     begin match question with
-      | { q_class = Q_IN; q_name; _ } when is_mdns q_name ->
-        Log.debug (fun f -> f "DNS lookup of %s: return NXDomain" (Dns.Name.to_string q_name));
+      | { q_class = Q_IN; q_name; _ } when q_name = localhost_local ->
+        Log.debug (fun f -> f "DNS lookup of localhost.local: return NXDomain");
         Lwt.return (q_name, [])
       | { q_class = Q_IN; q_type = Q_A; q_name; _ } ->
         getaddrinfo (Dns.Name.to_string q_name) Unix.PF_INET
--- a/src/hostnet/host_uwt.ml
+++ b/src/hostnet/host_uwt.ml
@@ -961,16 +961,13 @@ module Dns = struct
         | _ -> acc
       ) [] x
 
-  let is_mdns name =
-    match List.rev @@ Dns.Name.to_string_list name with
-    | [] -> false
-    | n :: _ -> n = "local"
+  let localhost_local = Dns.Name.of_string "localhost.local"
 
   let resolve_getaddrinfo question =
     let open Dns.Packet in
     begin match question with
-      | { q_class = Q_IN; q_name; _ } when is_mdns q_name ->
-        Log.debug (fun f -> f "DNS lookup of %s: return NXDomain" (Dns.Name.to_string q_name));
+      | { q_class = Q_IN; q_name; _ } when q_name = localhost_local ->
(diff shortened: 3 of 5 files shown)


christhomas:
Yes, in development, I use local domains as a way to have a real hostname and its logical and understandable, the software I work on uses fqdn's for accessing microservices and I cannot easily swap this for using docker dns by its container name or something.

so this is an absolute must, even if its slow, I need it, if this is opted out to get removed again or something in the future because of a performance problem, please consider adding an option to allow me to manually enable it and I will take the performance hit as opposed to the software just not working anymore, which is worse.

christhomas:
my understanding of this pull request is that if I can quickly switch my software to using localdomain, or development, or another tld, it will work? is that right? cause I need this to be solved today... :(

djs55:
Any other domain should work -- this was a special case for`.local` only. If changing the domain doesn't work then try a temporary downgrade to [17.03.1-ce-mac5 (16048)](https://download.docker.com/mac/stable/16048/Docker.dmg) (previous stable)

christhomas:
I actually just swapped out .local for .localdomain and in fact that doesnt work either, however a specific hostname that I put in my hosts folder, I can ping from inside and output the container without any problem

christhomas:
For example, here is a test run on whether it will match from the hosts or dns

test-hosts[0,1,2].me.localdomain are all valid from the host when pinging

```
me-chris-thomas:microservice me-20150103$ cat /etc/hosts
##
# Host Database
#
# localhost is used to configure the loopback interface
# when the system is booting.  Do not change this entry.
##
127.0.0.1	localhost
255.255.255.255	broadcasthost
::1             localhost

10.254.254.254 test-hosts1.me.local
10.254.254.254 test-hosts2.me.localdomain

me-chris-thomas:core-message me-20150103$ docker exec -it microservice_dev_phpfpm_1 ping test-hosts0.me.localdomain
ping: unknown host

me-chris-thomas:core-message me-20150103$ ping test-hosts0.me.localdomain
PING test-hosts0.me.localdomain (10.254.254.254): 56 data bytes
64 bytes from 10.254.254.254: icmp_seq=0 ttl=64 time=0.043 ms

me-chris-thomas:core-message me-20150103$ docker exec -it microservice_dev_phpfpm_1 ping test-hosts1.me.local
PING test-hosts1.me.local (10.254.254.254): 56 data bytes
64 bytes from 10.254.254.254: icmp_seq=0 ttl=37 time=0.192 ms
^C--- test-hosts1.me.local ping statistics ---
1 packets transmitted, 1 packets received, 0% packet loss
round-trip min/avg/max/stddev = 0.192/0.192/0.192/0.000 ms

me-chris-thomas:core-message me-20150103$ docker exec -it microservice_dev_phpfpm_1 ping test-hosts2.me.localdomain
PING test-hosts2.me.localdomain (10.254.254.254): 56 data bytes
64 bytes from 10.254.254.254: icmp_seq=0 ttl=37 time=0.247 ms
^C--- test-hosts2.me.localdomain ping statistics ---
1 packets transmitted, 1 packets received, 0% packet loss
round-trip min/avg/max/stddev = 0.247/0.247/0.247/0.000 ms
```

christhomas:
I have just tried crazy custom domains such as .monkey or .tiger and all of these work outside the container, but none of them work inside it, so this issue doesnt appear to be just about .local domains, it appears to be not attempting at all to use the host dns server, no matter what the domain name....

djs55:
@christhomas how are you configuring these extra domains? If they're in `/etc/hosts` then there might be a bug where the file isn't being re-read -- perhaps try restarting Docker for Mac to see if it makes a difference.

christhomas:
I am running a dns server in a container and I have added the domains there as wildcards, my browser works with them no problem, opens the software without a glitch, but the second it tries to access another piece of software it needs to resolve the hostname to an ip address, inside the container, it fails like above.

so seems that anything inside this dns server is not available, so docker is still not respecting my hosts dns configuration

christhomas:
I downgraded docker and it seems to be working, so I am going to keep with this version until you guys find another fix for this, seems there was a regression somewhere.

this version works with local, localdomain, monkey, so seems the problem came in quite recently

> Version 17.04.0-ce-rc2-mac6 (16165)
> Channel: edge
> 142462d3df

djs55:
@christhomas could you describe the host configuration in more detail so I can replicate the problem locally? How does the host know to use the DNS server in the container?

christhomas:
I have an ip alias for localhost to get around the machine 127.0.0.1 and container 127.0.0.1 problem thats just a simple: sudo ifconfig lo0 alias 10.254.254.254

I installed and configured a dns server inside a container, I setup a wildcard for "company.local$" as a regex pattern for matching any domain, I set the wildcard to respond with ip address: 10.254.254.254

I update my machine to use this dns server by adding its ip address to the nameservers using:
sudo networksetup -setdnsservers "Wi-Fi" 10.254.254.254 8.8.8.8 8.8.4.4

now pinging any company.local domain will resolve to 10.254.254.254 regardless of whether there is actually anything on that domain, its just a wildcard.

if I run any container, I should be able to also ping that same hostname, but I cannot do that from inside the container, only outside will the dns I installed locally respond with the ip address 10.254.254.254, inside the container I just get the responses I mentioned above, NXDOMAIN, which means no domain found

Issue #255, opened 2017-07-13
Invalid argument on pull 


1605200517:
Error log on a windows client:

```
[11:19:44.279][VpnKit         ][Error  ] vpnkit.exe: PPP.listen callback caught (Invalid_argument

[11:19:44.279][VpnKit         ][Info   ]   "Cstruct.blit src=[6296,1360](65536) dst=[0,4096](4096) dst-off=3673 `len=1360")
```

https://github.com/moby/moby/issues/34080

djs55:
The version of vpnkit in 17.06.0 is befe8be9b45f6c4d50cf7c8105bfdf101dfe8982 which is built against [cstruct.2.4.1](https://github.com/mirage/ocaml-cstruct/tree/v2.4.1).

The specific error `"Cstruct.blit src=[6296,1360](65536) dst=[0,4096](4096) dst-off=3673 len=1360")` means
- copy `1360` bytes
- from offset `6296` in `src` which has a total length of `65536`
- to offset `3673` in `dst` which has a total length of `4096` and therefore only `423` bytes remaining. Since `423` is less than `1360` there is a problem.

The `65536` buffer is probably the [read_buffer](https://github.com/moby/vpnkit/blob/d660726b4ea1b7b20d044b773bd4723aa68e81fe/src/hostnet/host_uwt.ml#L362) created for the outgoing connection to the registry. The `4096` buffer is probably an "I/O page" destined for transmission to the VM.

The fact that this bug only manifests on Windows strongly suggests it is in the Windows-specific code, which is currently
- support for jumbo ethernet frames; and
- [Hyper-V socket bindings](https://github.com/mirage/ocaml-hvsock/tree/v0.13.0).

Regarding jumbo ethernet frame support, [this function](https://github.com/djs55/mirage-tcpip/blob/f5b16043c57f71d5fdf1a7cf7ab7ccdde291cca1/tcp/user_buffer.ml#L164) on the TCP transmit path has a suspicious `Cstruct.blit`. If a large frame is being assembled out of small parts, the whole thing is not guaranteed to fit in one 4096 byte page. I'll prepare an experimental build without jumbo frame support to see if this fixes the issue.

djs55:
@1605200517 could you try an experiment? I'd like to see whether the problem is triggered by the large MTU setting on Windows.

[...]

PR #258, opened 2017-07-19
Temporarily revert the MTU to 1500

djs55:
This reverts 732bdaa734db1f94f0e5e859974dfaa320ad5587 and may prevent https://github.com/djs55/miragetcpip/blob/f5b16043c57f71d5fdf1a7cf7ab7ccdde291cca1/tcp/user_buffer.ml#L164 from raiseing `Invalid_argument` exceptions

Related to #255


djs55:
Merging since this did indeed fix the reported problem. Once the MTU-related issues in the tcp/ip stack are fixed then we can revert this revert.

PR #262, opened 2017-07-28
Mirage3

samoht:
This is built on top of the major cleanup that I've pushed in #259

Everything compiles but it seems that there are 3 missing functions in the tcpip stack:

- `Stack_ipv4.set_ip_netmask`
- `Stack_ipv4.set_ip_gateways`
- `Stack_tcp_wire.src_of_id` (e.g. there is no way to get the source IP)

Also the tests are compiling but not passing, either because of my very flaky connection or because I broke something with the `dns` ref that I modified to be `None` by default. I will investigate more next week.

PR #267, opened 2017-08-03
Fix dns dev version

samoht:
/cc @MagnusS 

Can you try to see if it works better for you?


MagnusS:
Hm it didn't work with just an update and a new --deps-only install, it still installed conduit 1.0.0. I'm trying to build on a new root now, in case my testing left it in a weird state - but takes a short while as it has to build ocaml 4.04.2 after the homebrew update

samoht:
Not sure if it is worth spending too much time on this anyway ... hopefully things will become easier once #262 is merged

MagnusS:
This patch + update upstream deps with `make` in `/repo` fixed it for me. I now have `conduit 0.14.5` and build instructions work again

MagnusS:
I also had to run `opam switch 4.04.2` first

samoht:
@MagnusS: `make OSS-LICENSE` is now failing

MagnusS:
Ok, as discussed on slack I'll try to just add the constraint to mirage-conduit manually instead of doing a full update to avoid conflicts with #262.

samoht:
The mirage3 branch has been merged, so I'm closing this PR.

PR #268, opened 2017-08-08
Remove non-bridge mode

MagnusS:
The virtual bridge support in vpnkit has been enabled by default in Docker for Mac for several edge releases and 17.06 stable without any reported issues. This PR removes the old code that could still be enabled with a database key.

Changes:
- Remove non-bridge mode everywhere and move the vnet switch to the config, so we don't have to pass it around
- Rename l2_switch and other l2_* variables to vnet_* to make it clearer that they relate to the virtual network
- Fix tests to use a dynamic mac and run them on the new bridge (bridge mode was disabled for tests)

### DIFF
--- a/src/bin/main.ml
+++ b/src/bin/main.ml
@@ -322,7 +322,7 @@ let hvsock_addr_of_uri ~default_serviceid uri =
 
     let uri = Uri.of_string socket_url in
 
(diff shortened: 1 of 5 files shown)


djs55:
Looks good to me -- merging since the CI is green.

PR #269, opened 2017-08-08
Add a Endpoint.destroy which closes active connections

djs55:
Previously there was no way to locate the connections associated with
an endpoint to shut them down. This patch adds a map of TCP `id` to
`unit Lwt.u` and a function `Endpoint.destroy` which triggers the
disconnection of all the active connections.

Related to #260


djs55:
The purpose of this PR is to run the CI and build a binary we can experiment with.

Issue #209, opened 2017-04-10
Move preferred IP into new protocol command


MagnusS:
The vmnet protocol used by vpnkit and hyperkit should have a `preferred_ip` command to request a specific IP with the uuid. See also https://github.com/docker/vpnkit/pull/207#discussion_r110441704

Changes needed:
* [x] Vpnkit: protocol needs to be extended to include a [new command](https://github.com/docker/vpnkit/blob/master/src/hostnet/vmnet.mli#L48)
* [x] Hyperkit: accept a `preferred_ip=` option in `pci_virtio_net_vpnkit` and send the new command to vpnkit if present

MagnusS:
I'm working on adding a new `Preferred_ipv4` command to `vpnkit` now. Since the first packet with the command has a fixed size and only has room for the UUID, I'll have to extend it with 4 bytes for the IP. This will not be backwards compatible and requires a protocol version bump. 

Alternatively, I could try to keep the code compatible by making the buffer length dynamic -- but unless bumping the version will be problematic I'm not sure it's worth the restructuring this would require.

djs55:
There aren't very many other implementations of this protocol. The only other one is in `com.docker.vmnetd` in D4M which we can update easily enough. Will it be easy to probe the version in a way that will work with both old and new versions? That's all we need to trigger an automatic upgrade. Ideally we'd keep the version exchange messages the same format to keep this easy.

MagnusS:
Fixed in https://github.com/moby/vpnkit/pull/277 and https://github.com/moby/hyperkit/pull/154, so closing this issue.

PR #277, opened 2017-08-25
Protocol update: Add support for error messages, preferred IPv4 addresses

MagnusS:
This updates the protocol from v1 to v22. The new features are:

- New Preferred_ipv4 command may be used to request a specific IP which will be assigned to the UUID. The UUID can then be reused to get the same IP in the future. This replaces the old mechanism that encoded the requested IP in the UUID itself.
- New Response message returned for Ethernet and Preferred_ipv4 commands. The Response message either contains a Vif structure on success or a Disconnect with a reason message on failure. This allows errors to be reported back to the client, such as when the preferred IP address is already in use.

This PR also adds test for the bridge and the Preferred_ipv4 command. See individual commits for more details.

These changes are not backwards compatible. Other libraries that use VPNKit have to be updated or they will be disconnected by the server. I'll make PRs to Hyperkit, Linuxkit and Docker for Desktop when this passes initial review.

Related to #209 and linuxkit/linuxkit#1970

### DIFF
--- a/src/hostnet/vmnet.ml
+++ b/src/hostnet/vmnet.ml
@@ -273,6 +273,25 @@ module Make(C: Sig.CONN) = struct
   let client_log_prefix = "Vmnet.Client"
 
   let server_negotiate ~fd ~connect_client_fn ~mtu =
+    let assign_uuid_ip uuid ip =
+      connect_client_fn uuid ip >>= fun mac ->
+      match mac with
+      | Error (`Msg msg) ->
+          let buf = Cstruct.create Response.sizeof in
+          let (_: Cstruct.t) = Response.marshal (Disconnect msg) buf in
+          Log.err (fun f -> f "%s.negotiate: disconnecting client, reason: %s" server_log_prefix msg);
+          Channel.write_buffer fd buf;
+          with_flush (Channel.flush fd) @@ fun () ->
+          failf "%s.negotiate: disconnecting client, reason: %s " server_log_prefix msg
+      | Ok client_macaddr -> 
+          let vif = Vif.create client_macaddr mtu () in
+          let buf = Cstruct.create Response.sizeof in
+          let (_: Cstruct.t) = Response.marshal (Vif vif) buf in
+          Log.info (fun f -> f "%s.negotiate: sending %s" server_log_prefix (Vif.to_string vif));
+          Channel.write_buffer fd buf;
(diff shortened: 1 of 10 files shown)


src/hostnet/vmnet.ml
```diff
@@ -25,7 +25,7 @@ module Init = struct
 
   let default = {
     magic = "VMN3T";
-    version = 1l;
+    version = 22l;
```
samoht: is 22 expected? shouldn't be just 2?

MagnusS: Yes, it is to be higher than the protocol version used internally in Docker for Mac, which is currently 21.

djs55:
Looks good to me!

One request: when you make a PR for hyperkit, could you try to make it connect with the new protocol version first, and then fall back to the old one? I'd like to make sure that `linuxkit` users can run against either D4M 17.06.0 (with the old protocol) or a later version (TBD) (with the new protocol)

MagnusS:
@djs55 Thanks for the review. I can make the hyperkit vpnkit driver backwards compatible. The IPv4 address is passed as a new option now though, so `linuxkit run --ip [...]` will probably still fail on an older version with `invalid option: preferred_ipv4` unless you update both. But you would get regular network support at least.

djs55:
OK sounds fair enough -- if you request the new feature you need the new version. Completely reasonable :)

MagnusS:
Hm, but thinking about this a bit more I'm not sure we have to make the hyperkit driver backwards compatible for linuxkit. If you update Docker for Mac you'll get the new version of both `vpnkit` and `hyperkit`, so they should be in sync.  `linuxkit run` just uses the Go API, which is mostly the same, so all linuxkit/hyperkit combinations should work (unless you use `--ip`).

djs55:
Ah good point -- I forgot `linuxkit` was using the D4M distribution of `hyperkit` rather than compiling its own. It sounds like we don't need to do anything apart from ensure that D4M ships a coherent set of versions then.

djs55:
BTW feel free to merge this whenever is convenient for you!

MagnusS:
@djs55 thanks - I think I'll leave this (and related PRs) open until it passes tests on D4M/W.

MagnusS:
Tests passed in D4D, so merging!

PR #278, opened 2017-09-06
Enable TCP keep-alives on external connections

djs55:
We wish to detect when connections have disappeared (e.g. when the
host network environment changes) and free any associated filehandles.

### DIFF
--- a/src/hostnet/host.ml
+++ b/src/hostnet/host.ml
@@ -424,6 +424,12 @@ module Sockets = struct
           Lwt.catch (fun () ->
               let sockaddr = make_sockaddr (ip, port) in
               Uwt.Tcp.connect fd ~addr:sockaddr >>= fun () ->
+              let error = Uwt.Tcp.enable_keepalive fd 1 in
+              if Uwt.Int_result.is_error error then begin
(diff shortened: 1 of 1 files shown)


hablutzel1:
I might be wrong, but the changes seem to be setting the TCP keepalive initial delay to 1 second (see https://fdopen.github.io/uwt/Uwt.Tcp.html#VALenable_keepalive). Isn't it excessively low considering that the default value uses to be much larger, e.g. 7200 in Linux systems as mentioned in https://tldp.org/HOWTO/TCP-Keepalive-HOWTO/usingkeepalive.html.

waynestevensinsight:
I agree with @hablutzel1, this keepalive initial delay and the subsequent intervals (both 1s) seem very low. In fact, setting these values at all can (and does) interfere with application-level settings. I have an application that explicitly sets these values at the socket level. After getting inconsistent results when running in Docker Desktop versus on native Linux, I had to use Wireshark and tcpdump to discover these 1s keepalives. This effectively means I can't use Docker Desktop for developing/testing this functionality of my application, as there is no way to turn these off - would it be possible to expose an an option for this?

PR #281, opened 2017-09-07
Add vpnkit-tap-vsockd

MagnusS:
This adds `vpnkit-tap-vsockd` (previously `tap-vsockd`) from linuxkit. This component is needed to provide vpnkit networking in Docker for Windows.

I've extracted original history (except various build scripts) and added a new Makefile and Dockerfile. I also added a commit to update to the current vpnkit version (22).

I wasn't sure where to put it, so I made `/c` (as with `/go`)


MagnusS:
@djs55 Thanks. Yes, I think it could be useful with a hub build here to simplify testing until we have a signed package in linuxkit. The Dockerfile/Makefile may require some modifications though, not sure

MagnusS:
Merging this for now. It would be nice to have some tests for this, but not sure if we can in the current CI - but at least we could try to build it

Issue #286, opened 2017-09-15
Failure to start on some Windows machines


djs55:
A small number users report that `vpnkit` won't start unless they run it as `Administrator`. The problem was introduced somewhere in #262 which is unfortunately a large PR. We need to generate some test builds with aspects of it reverted to try to narrow down the cause.

djs55:
Adding some debug on a machine with a repro highlights the user-space ICMP socket -- on the buggy system the `socket` call fails with `EPERM`.

PR #287, opened 2017-09-15
Do not merge: test a downgrade of libuv to 1.12.0

djs55:
In the working build we used `uwt.0.2.0` with `libuv.1.11.0`. In the bad build we used `uwt.0.2.1` with `libuv.1.12.0`

These versions are all old so we ought to upgrade soon.

Related to #286


djs55:
The bug is still there.

PR #288, opened 2017-09-15
Remove unnecessary Unix linkage

djs55:
Previously strange things happened when we linked both `uwt` and `lwt.unix`, `lwt.preemptive` at the same time. Obviously the `lwt.unix` functions won't work if used, but I suspect there may be module-level initialisation that conflicts (or could conflict) with `uwt`. Some extra linkage happened during the recent backend refactor *and* the versions of `uwt` and `lwt` were both upgraded.

Related to #286 but this is a good change anyway.


djs55:
FWIW this artefact https://ci.appveyor.com/api/buildjobs/jwsn3wjy9jkk7wo6/artifacts/vpnkit.exe doesn't fix the bug in #286

PR #291, opened 2017-09-19
go: add a client for the Vmnet protocol

### LINKED ISSUE #257: Support for port forwarding from host to guest
I'm working with [LinuxKit](https://github.com/linuxkit/linuxkit) and would like to be able to support port forwarding with the Hyperkit+Vpnkit backend on MacOS in the same way as the QEMU backend works.

With the QEMU backend I can do:
```
linuxkit run qemu -publish 1234:22 «image»
```
which maps `localhost:1234` to port `22` on the VMs IP address, with that I can `ssh -p 1234 localhost` to reach an ssh server running in the VM. This essentially uses the `hostfwd` argument which QEMU supports with the `-net` option.

Note that there is no need for any in VM process other than sshd itself since this does TCP forwarding to the VM over the Ethernet link, targeting the IP address which QEMU knows because it assigned it.

I'm trying to make
```
linuxkit run hyperkit -publish 1234:22 «image»
```
behave the same such that `linuxkit run -publish 1234:22 «image»` works regardless of the host OS (on Linux this will autodetect QEMU, on MacOS it will use hyperkit)

I began adding support using the `github.com/moby/vpnkit/go/pkg/vpnkit` library to talk to the vpnkit instance in order to ask it to perform the forwarding (using the [`vpnkit-expose-port`](https://github.com/moby/vpnkit/blob/master/go/cmd/vpnkit-expose-port/main.go) code for reference).

I ran into a roadblock however which is that I do not know the guest IP address in order to make the request. I'm also lead to believe that the current vpnkit forwarding arrangements require an agent with the VM which I'd like to avoid.

So I (think I) have two asks from VPNKit so I can build this:

* Provide a mechanism to retrieve the IP address dynamically assigned to the VM, perhaps based off the VPNKit UUID? Note that VPNKit supports static IP address allocation based on a magic syntax for the UUID (where the top 4 bytes encode the IP and the other bytes are zero). I want to avoid having to statically assign IP addresses though.
* Ensure that the port forwarding can work without the need for an in guest agent.

My incomplete code (which doesn't compile since I don't have a guest IP address to use) can be found [in my tree](https://github.com/ijc/linuxkit/tree/hyperkit-port-forwarding)

djs55:
This will allow us to connect, allocate UUIDs and query the IP mapping.

The API is demonstrated using an example command `vpnkit-register-uuid`

Note we query the IP mapping via issuing a DHCP request, like a real client would.

Fixes #257 -- to forward from host to guest without an agent:

1. generate a UUID for the VIF (cache this somewhere if you want the IP to persist)
2. use the new vmnet API to connect a VIF and query the IP
3. use the UUID when starting hyperkit
4. use the port forwarding API to forward to the guest IP

### DIFF
--- a/go/pkg/vpnkit/vmnet.go
+++ b/go/pkg/vpnkit/vmnet.go
@@ -1,0 +1,588 @@
+package vpnkit
+
+import (
+	"bytes"
+	"context"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"os"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+// Vmnet describes a "vmnet protocol" connection which allows ethernet frames to be
+// sent to and received by vpnkit.
+type Vmnet struct {
+	conn          net.Conn
+	remoteVersion *InitMessage
+}
+
+// NewVmnet constructs an instance of Vmnet.
+func NewVmnet(ctx context.Context, path string) (*Vmnet, error) {
+	d := &net.Dialer{}
+	conn, err := d.DialContext(ctx, "unix", path)
+	if err != nil {
+		return nil, err
+	}
+	var remoteVersion *InitMessage
+	vmnet := &Vmnet{conn, remoteVersion}
+	err = vmnet.negotiate()
+	if err != nil {
+		return nil, err
+	}
+	return vmnet, err
+}
+
+// Close closes the connection.
+func (v *Vmnet) Close() error {
+	return v.conn.Close()
+}
+
+// InitMessage is used for the initial version exchange
+type InitMessage struct {
+	magic   [5]byte
+	version uint32
+	commit  [40]byte
+}
+
+// String returns a human-readable string.
+func (m *InitMessage) String() string {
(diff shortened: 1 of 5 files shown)


go/pkg/vpnkit/vmnet.go
```diff
@@ -0,0 +1,588 @@
...
+	}
+	ClientMAC := mac[:]
+	conn := v.conn
+	var IP net.IP
+	return &Vif{MTU, MaxPacketSize, ClientMAC, IP, conn}, nil
+}
+
+// ConnectVif returns a connected network interface with the given uuid
+func (v *Vmnet) ConnectVif(uuid uuid.UUID) (*Vif, error) {
+	e := Ethernet{uuid}
```
MagnusS: Would it make sense to support `Preferred_ipv4` here too? The IP should be known, but the command could fail from server with IP in use or invalid IP.

djs55:
@MagnusS thanks -- it's gone green so I'll merge.

I'll have a quick look at supporting `Preferred_ipv4` -- should be a small extension.

PR #292, opened 2017-09-19
go: add binding for Preferred_ipv4

djs55:
This allows Go programs to check that a UUID <-> IP mapping exists and to create one (provided the UUID and IP are free)


djs55:
FYI @MagnusS

PR #293, opened 2017-09-22
Add /bin/mount to vpnkit-9pmount-vsock container

MagnusS:
The container was built from scratch, but vpnkit-9pmount-vsock calls
/bin/mount to set up the mount point after opening the fd.

Also allow networking in container build.


MagnusS:
CI is green, so merging to trigger cloud build

PR #295, opened 2017-10-04
[do not merge] Remove a potentially dodgy C stub

djs55:
There have been some observed crashes which might implicate this stub. This PR will build a version without this function.


djs55:
The crash was actually caused by libgmp so this PR can be closed.

PR #296, opened 2017-10-04
[do not merge] Isolate a crash by removing the dependency on libgmp

djs55:


djs55:
Dependency on libgmp has been removed.

PR #305, opened 2017-11-07
go: use uint16 for ports

guillaumerose:
Using int16 caused large ports to be negative.

For example, the original port was 44134.
```
$ ./vpnkit-list-ports
2017/11/07 11:26:02 Dialling unix /Users/guillaumerose/Library/Containers/com.docker.docker/Data/s51
tcp forward from 0.0.0.0:-21402 to 10.111.24.47:-21402
```

### DIFF
--- a/go/pkg/vpnkit/port.go
+++ b/go/pkg/vpnkit/port.go
@@ -18,14 +18,14 @@ type Port struct {
 	client  *datakit.Client
 	proto   string
 	outIP   net.IP
-	outPort int16
+	outPort uint16
 	inIP    net.IP
-	inPort  int16
+	inPort  uint16
 	handle  *datakit.File
 }
 
 // NewPort constructs an instance of Port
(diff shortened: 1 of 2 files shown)


dsheets:
Why not `uint16`?

guillaumerose:
Looking at kubernetes project, they use int32. I thought that was a good reference.

https://github.com/kubernetes/api/blob/master/core/v1/types.go#L1636

dsheets:
The [0,1<<16) range is not tested or enforced.

The Kubernetes source indicates that the value of the field must be within the range of `uint16`. In fact, depending on the use case, the Kubernetes comment may be wrong as `0 < x < 65536` does not include 0 which is a reserved but valid binding port which indicates OS port selection.

guillaumerose:
I changed the type to uint16

djs55:
@guillaumerose thanks! Don't worry about the CircleCI or appveyor breakage -- they're the OCaml build only (it looks like I removed a branch in another repo that was still being referenced here).

PR #309, opened 2017-11-15
proxy: move helper methods from go/cmd/vpnkit-forwarder to go/pkg/libproxy

errordeveloper:
Signed-off-by: Ilya Dmitrichenko <errordeveloper@gmail.com>

### DIFF
--- a/go/cmd/vpnkit-forwarder/one.go
+++ b/go/cmd/vpnkit-forwarder/one.go
@@ -1,13 +1,7 @@
 package main
 
 import (
-	"errors"
-	"fmt"
-	"log"
-	"net"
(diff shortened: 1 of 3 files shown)


errordeveloper:
Are CircleCI and AppVeyor supposed to pass?

errordeveloper:
CircleCI failure seems unrelated to Go code, it's the OCaml part that fails there, as far as I can tell.

djs55:
CircleCI and appveyor are supposed to pass :) but don't worry -- it must be the OCaml code. I'll fix it separately. Only travis builds the Go code I think.

LGTM!

Issue #307, opened 2017-11-10
go: vpnkit.Connection should have a Close()


djs55:
This is needed for cleanup. I think the protocol requires us to keep the physical connection open because the `fids` are tied to it and the resources are tied to the `fids`, but a future non-9P implementation should support a better API... so we could think about being future proof.

See https://github.com/linuxkit/linuxkit/pull/2701#discussion_r150187924

Issue #306, opened 2017-11-07
vpnkit-expose-port-controller could call the API directly instead of exec


errordeveloper:
We currently have some code that execs `vpnkit-expose-port` and stores pid files in state dir etc, we shouldn't need to do all that if we could call the Go API directly.

PR #310, opened 2017-11-16
proxy: move helper methods for muliplexed connections from go/cmd/vpnkit-forwarder to go/pkg/libproxy

errordeveloper:
@djs55 PTAL

### DIFF
--- a/go/cmd/vpnkit-forwarder/many.go
+++ b/go/cmd/vpnkit-forwarder/many.go
@@ -15,43 +15,42 @@ func manyPorts() {
 	var (
 		vsockPort = flag.Int("vsockPort", 62373, "virtio-vsock port")
 		hvGUID    = flag.String("hvGuid", "0B95756A-9985-48AD-9470-78E060895BE7", "Hyper-V service GUID")
+		listener  net.Listener
 	)
 	flag.Parse()
-
-	listeners := make([]net.Listener, 0)
 
 	vsock, err := vsock.Listen(vsock.CIDAny, uint32(*vsockPort))
 	if err != nil {
 		log.Printf("Failed to bind to vsock port %d: %#v", vsockPort, err)
 	} else {
-		listeners = append(listeners, vsock)
+		listener = vsock
 	}
 	svcid, _ := hvsock.GUIDFromString(*hvGUID)
 	hvsock, err := hvsock.Listen(hvsock.HypervAddr{VMID: hvsock.GUIDWildcard, ServiceID: svcid})
 	if err != nil {
 		log.Printf("Failed to bind hvsock guid: %s: %#v", *hvGUID, err)
 	} else {
-		listeners = append(listeners, hvsock)
+		listener = hvsock
+	}
+
+	if listener == nil {
+		log.Fatal("Failed to bind vsock or hvsock")
 	}
 
 	quit := make(chan struct{})
 	defer close(quit)
 
-	for _, l := range listeners {
-		go func(l net.Listener) {
-			for {
-				conn, err := l.Accept()
-				if err != nil {
-					log.Printf("Error accepting connection: %#v", err)
-					return // no more listening
-				}
-				if err := libproxy.HandleMultiplexedConnections(conn, quit); err != nil {
-					log.Println(err)
-					conn.Close()
-				}
+	for {
+		conn, err := listener.Accept()
+		if err != nil {
+			log.Printf("Error accepting connection: %#v", err)
+			return // no more listening
+		}
+		go func() {
+			defer conn.Close()
+			if err := libproxy.HandleMultiplexedConnections(conn, quit); err != nil {
+				log.Println(err)
 			}
-		}(l)
+		}()
 	}
-	forever := make(chan struct{})
-	<-forever
 }
--- a/go/cmd/vpnkit-forwarder/many.go
+++ b/go/cmd/vpnkit-forwarder/many.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"encoding/binary"
 	"flag"
 	"log"
 	"net"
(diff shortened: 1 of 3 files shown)


go/cmd/vpnkit-forwarder/many.go
```diff
@@ -46,72 +45,13 @@ func manyPorts() {
...
-						break
-					default:
-						log.Printf("Unknown protocol: %d", d.Proto)
-						conn.Close()
-						return
-					}
-				}(conn)
+				if err := libproxy.HandleMultiplexedConnections(conn, quit); err != nil {
+					log.Println(err)
+					conn.Close()
```
errordeveloper: From what I can see, we only close on errors, but I wonder whether we should have `defer conn.Close()`, but I don't know enough about this.

go/cmd/vpnkit-forwarder/many.go
```diff
@@ -46,72 +45,13 @@ func manyPorts() {
...
+				}
 			}
 		}(l)
 	}
 	forever := make(chan int)
```
errordeveloper: I think we should use `chan struct{}` here.

go/cmd/vpnkit-forwarder/many.go
```diff
@@ -46,72 +45,13 @@ func manyPorts() {
...
-						}
-						proxy.Run()
-						break
-					default:
-						log.Printf("Unknown protocol: %d", d.Proto)
-						conn.Close()
-						return
-					}
-				}(conn)
+				if err := libproxy.HandleMultiplexedConnections(conn, quit); err != nil {
```
errordeveloper: `quit` doesn't have to be `chan bool` (value is never used and it can be `chan struct{}` instead), but I'm not sure if breaking the API is worse it.

go/cmd/vpnkit-forwarder/many.go
djs55: It's not obvious (which is bad) but I think `HandleMultiplexedConnections` calls `HandleTCPConnection` which calls `Close` https://github.com/moby/vpnkit/blob/e1e5e6b8a0b0dea4e82dd5d23f4f8eb49f5ec69c/go/pkg/libproxy/tcp_proxy.go#L81 and `UDPProxy.Run` which ... looks like it might leak the listening socket? It's possible we didn't notice because in the one-process-per-port model it would be cleaned up anyway when the process quits.

I think `HandleTCPConnection` should probably not call `client.Close()`, since `client` is passed in from the caller. I think it would be clearer if the caller retained ownership of the connection and called `defer conn.Close()` as you suggest.

go/cmd/vpnkit-forwarder/many.go
djs55: Sounds good!

go/cmd/vpnkit-forwarder/many.go
djs55: I don't mind breaking this API :)

errordeveloper:
I'm happy to fix a few more things in this PR. I've added a commit wich changes type of chan, but will hold on with changes to how the connection is closed, as I'm not exactly sure at this time of the day – will have to take a fresh look in the morning.

Also, how should I verify my changes work – is there a test for this?

go/cmd/vpnkit-forwarder/many.go
errordeveloper: Closing on error case is obvious from outside the library call, but I'm not so sure how to handle the two non-error cases of `client.Close()` in `HandleTCPConnection`...

go/cmd/vpnkit-forwarder/many.go
errordeveloper: Good spirit!

go/cmd/vpnkit-forwarder/many.go
```diff
@@ -46,72 +45,13 @@ func manyPorts() {
...
-						}
-						proxy.Run()
-						break
-					default:
-						log.Printf("Unknown protocol: %d", d.Proto)
-						conn.Close()
-						return
-					}
-				}(conn)
+				defer conn.Close()
```
errordeveloper: This isn't right though, cause now we aren't closing in non-error case, to fix that we could add another goroutine with an explicit return, but I'm not too sure what is the intention here.

To be honest, it seems like this would become easier if we get rid of the `forever` channel and have single listener variable instead of a slice, what do you think?

errordeveloper:
@djs55 I just noticed https://github.com/moby/vpnkit/issues/307, which seems to be about the host end, but wanted to double check with you.

go/cmd/vpnkit-forwarder/many.go
djs55: I agree -- since there is only one working listener (vsock or hvsock but not both) we should simplify.

djs55:
@errordeveloper yeah #307 is on the host (Unix domain socket) end. @rn observed in his recent linuxkit work that it wasn't possible to clean up the control connection so it leaked.

errordeveloper:
I got rid of the listener slice and simplified the main loop.

errordeveloper:
> Also, how should I verify my changes work – is there a test for this?

@djs55 should I take any steps or this is covered in CI somehow?

djs55:
The `libproxy` code has some old test cases but they aren't running from the CI at the moment (see https://github.com/moby/vpnkit/blob/master/go/pkg/libproxy/network_proxy_test.go). (The `libproxy` package came originally from `docker/docker`, now `moby/moby`). I think we should probably resurrect this and and extend it to cover more stuff. It's probably bit-rotted :(

The control interface stuff (Unix domain socket and 9P mount) are much harder to test. I think we should start with the `libproxy` proxying code.

errordeveloper:
The test cases in `go/pkg/libproxy/network_proxy_test.go` use `NewProxy`, which we don't have anymore. I'm not exactly keen to rewrite those test, as it seems like the purpose of the code has evolved a little (feel free to convince me otherwise), but at this point I'd rather run through some kind of end-to-end/integration test where all other bits in place – e.g. LinuxKit VM.

An easier way to test some of this would be to actually implement #306, however that won't cover `HandleMultiplexedConnections`, so I wonder what uses that functionality...?

errordeveloper:
So basically, I'm not very much concerned about changes made in #309 and trivial parts of this PR, but because I don't understand enough of about the outer context of `go/cmd/vpnkit-forwarder/many.go`, I think I'd be great to verify the changes I made to that code.

However, if you are quite happy with these changes overall, we can merge of course and find breakages when this gets integrated downstream, as I assume there exists some kind of test suite for the final product, such as d4m master or edge build.

djs55:
Yeah indeed there are tests downstream. That's fine for now!

djs55:
BTW travis is currently failing with
```
go build -o build/vpnkit-forwarder.linux --ldflags '-s -w -extldflags "-static"' --buildmode pie \
cmd/vpnkit-forwarder/proxy.go cmd/vpnkit-forwarder/main.go cmd/vpnkit-forwarder/one.go cmd/vpnkit-forwarder/many.go
# command-line-arguments
cmd/vpnkit-forwarder/many.go:44: listener.Accept undefined (type *net.Listener is pointer to interface, not interface)
make: *** [Makefile:23: build/vpnkit-forwarder.linux] Error 2
The command '/bin/sh -c GOPATH=/go make build/vpnkit-forwarder.linux' returned a non-zero code: 2
make: *** [build-forwarder-in-container] Error 2
```

errordeveloper:
Ok, we don't need a pointer, empty `net.Listener` is `nil`.

djs55:
Looks good to me!

PR #313, opened 2017-11-22
[do not merge] Restart on database key change

djs55:
Although the database using code will soon be removed, the transition
is easier if the behaviour is still the same in the meantime. This
patch restores the previous behaviour of restarting if certain keys
change.


djs55:
Some of the Windows end-to-end tests are failing so I'm reverting some of the Windows-specific changes to see if they're the cause.

djs55:
Windows build is failing with
```
# File "cohttp/src/link.ml", line 123, characters 10-11:
# Error: Multiple definition of the module name Fields.
#        Names must be unique in a given structure or signature.
```

djs55:
I don't think this is needed.

PR #318, opened 2017-12-13
make Kubernetes controller call libproxy directly instead of exec

### LINKED ISSUE #306: vpnkit-expose-port-controller could call the API directly instead of exec

errordeveloper:
Close #306.


errordeveloper:
@djs55 I've not tested this properly yet, so it's just WIP/FIY for now, and – no rush to review it.

djs55:
I've force pushed a rebase to latest master which has a travis fix.

errordeveloper:
@djs55 thanks! I've rebased this also.

djs55:
Thanks!

PR #325, opened 2018-01-08
HTTP: HEAD responses must not have bodies

djs55:
According to RFC2616 Section 9.4[1]

> The HEAD method is identical to GET except that the server
> MUST NOT return a message-body in the response.

Unfortunately the cohttp Response.has_body function doesn't take this
into account. This patch special-cases HEAD requests to work around this
issue.

[1] https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html


src/hostnet/hostnet_http.ml
```diff
@@ -282,10 +282,22 @@ module Make
                     (Cohttp.Response.sexp_of_t res)));
       let reader = Outgoing.Response.make_body_reader res outgoing in
       Incoming.Response.write ~flush:true (fun writer ->
-          match Incoming.Response.has_body res with
-          | `Yes     -> proxy_body_response ~reader ~writer
-          | `No      -> Lwt.return_unit
-          | `Unknown ->
+          match Cohttp.Request.meth req, Incoming.Response.has_body res with
+          | `HEAD, _ ->
```
ijc: Should this be `HEAD, Yes`? Since `HEAD, No`, is the expected state, isn't it?

djs55: Yeah that would be clearer -- so in that case the first clause would be specific to the bug and match the comment more closely. I'll change that.

PR #326, opened 2018-01-08
Backport the HTTP HEAD fix onto the stable branch

djs55:
- backport #325 
- add a mirror of the `menhir` package as the primary website is down


djs55:
Looks like all the tests have passed, so merging.

PR #331, opened 2018-01-09
Allow to create connections for arbitrary clients

pgayvallet:
In Docker4Windows, we need to create connections using hvsock datakit client instead of linux sockets. 

PR add a new `vpnkit.Connection` constructor function allowing to create connections for arbitrary datakit.Client


djs55:
The Windows CI might be stuck, but it's irrelevant anyway since it doesn't build the Go code. The travis CI does build the Go code and it's green.

This patch looks fine to me. Thanks!

PR #335, opened 2018-01-10
[stable] HTTP proxy: include the port number in the host header

djs55:
RFC2616[1] says that if the port number is not 80 (the default) then
the port must be included in the `Host` header.

[1] https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23

Related to https://github.com/docker/for-mac/issues/2392


djs55:
Looks like all the tests have passed.

PR #345, opened 2018-01-18
[stable 17.12.x] Backport HTTP proxy fixes

djs55:
#330 test: check that the HTTP proxy doesn't duplicate host: headers
#332 test: check the HTTP proxy handles HEAD requests
#334 test: strengthen the HTTP proxy host header test
#341 Fix HTTP proxy authentication
#343 allow excludes to contain IPs as well as IP prefixes
#344 dns: return NXDomain for a builtin where we don't have RRs


djs55:
Appveyor failed to download cygwin. Unfortunately there is no "rebuild" button. I'll merge speculatively and see if the merge build succeeds. .

PR #347, opened 2018-01-19
HTTP proxy: fix localhost

djs55:
HTTP traffic on port 80 or 443 are dispatched based on destination IP address:
 
- if to "localhost" (e.g. `docker.for.mac.localhost`) they are forwarded to 127.0.0.1
- if to anywhere else they are funnelled through the transparent proxy
    
HTTP traffic sent to vpnkit's gateway IP on port 3128 (the HTTP proxy) but whose final destination is one of the names for the host's localhost is unfortunately routed to the wrong place.

This patch adds a `localhost_names` argument to the HTTP proxy `route` function. If the destination host is one of the `localhost_names` then it is rewritten as `localhost` which will then cause connections to the host's `localhost` as expected.

### DIFF
--- a/src/hostnet/hostnet_http.ml
+++ b/src/hostnet/hostnet_http.ml
@@ -483,7 +483,7 @@ module Make
       let port = match Uri.port uri with None -> 80 | Some p -> p in
       Ok (host, port)
 
-  let route proxy exclude req =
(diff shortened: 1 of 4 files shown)


djs55:
Related to [docker/for-win#1130]

PR #357, opened 2018-02-12
go: set the default control socket path depending on OS

djs55:
Darwin and Windows use a different control socket path, so adopt the
convention that if the user doesn't specifiy a path override, we use
the system one.


djs55:
This code is built by travis (which has passed) but not by appveyor or circleCI.

PR #367, opened 2018-03-07
test: check the proxy respects Connection: close from origin servers

djs55:
This test starts an origin server on localhost which will respond with a minimal HTTP/1.0 response, with no connection-length or transfer-encoding. A client connects to the builtin HTTP proxy on port 3128 and requests a URI on the origin server. The test requires the body to be proxied correctly by the proxy (i.e. all data up to EOF should be transferred)

Tests #361

This PR is built on top of #366, only the last patch is relevant.

### DIFF
--- a/src/hostnet_test/test_http.ml
+++ b/src/hostnet_test/test_http.ml
@@ -745,6 +745,90 @@ let test_http_connect_tunnel proxy () =
         )
     end
 
+  let test_connection_close () =
+    let body = "Hello\n" in
+    Host.Main.run begin
+    Slirp_stack.with_stack ~pcap:"test_connection_close.pcap" (fun _ stack ->
+        with_server (fun flow ->
+            let ic = Incoming.C.create flow in
+            Incoming.Request.read ic >>= function
+            | `Eof ->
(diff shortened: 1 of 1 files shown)


src/hostnet_test/test_http.ml
```diff
@@ -745,6 +745,90 @@ let test_http_connect_tunnel proxy () =
...
+            let host = "127.0.0.1" in
+            let port = origin_server.Server.port in
+            Log.info (fun f -> f "HTTP origin server is on %s:%d" host port);
+            let open Slirp_stack in
+            (* Disable the proxy so the builtin proxy will have to fetch from the origin server *)
+            Slirp_stack.Debug.update_http ()
+            >>= function
+            | Error (`Msg m) -> failwith ("Failed to disable HTTP proxy: " ^ m)
+            | Ok () ->
+              (* Connect to the builtin HTTP roxy *)
```
akimd: The comment "Proxy" was for this line.

djs55:
Thanks, I fixed the "roxy" typo and used the `"%a"` format string as suggested.

PR #368, opened 2018-03-08
go: add a missing import

djs55:
The CI is not building the _darwin.go files.


djs55:
It looks like Appveyor failed for a spurious reason. Since appveyor doesn't build this code at the moment, I don't think this matters.

AkihiroSuda:
Seems closable now?

PR #373, opened 2018-04-06
Add a simple NTP server

djs55:
This replaces the UDP forwarder on the gateway address port 123 with a Simple NTP server which uses the host's wall-clock as a time source. This allows the VMs to synchronise with the host, even when the host itself is not synchronised anywhere sensible.

Related to [docker/for-mac#2076]

### DIFF
--- a/src/hostnet/ntp_server.ml
+++ b/src/hostnet/ntp_server.ml
@@ -8,19 +8,57 @@ module Log = (val Logs.src_log src : Logs.LOG)
 module Make
     (Ip: Mirage_protocols_lwt.IPV4)
     (Udp:Mirage_protocols_lwt.UDPV4)
-    (Clock: Mirage_clock_lwt.MCLOCK) =
+    (Clock: Mirage_clock_lwt.PCLOCK) =
 struct
 
+  (* From RFC2030 LOCL is an
+     > uncalibrated local clock used as a primary reference for
+     > a subnet without external means of synchronization *)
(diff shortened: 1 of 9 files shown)


akimd:
What's the status of this?  Didn't you already install it in D4D?

src/hostnet/ntp_server.ml
```diff
@@ -0,0 +1,26 @@
+let src =
```
akimd: > this is incorrect behaviour, but at least it type checks

:) :) :)

src/hostnet/ntp_server.ml
```diff
@@ -8,19 +8,57 @@ module Log = (val Logs.src_log src : Logs.LOG)
...
+    Cstruct.blit_from_string "LOCL" 0 buf 0 4;
+    Cstruct.BE.get_uint32 buf 0
 
-  let reply request = request
+  (* Convert a Mirage PCLOCK (days, picoseconds) into an NTP timestamp *)
+  let timestamp (days, picoseconds) =
+    (* Posix time starts at   1970-01-01 00:00:00 UTC
+       but NTP time starts at 1900-01-01 00:00:00 UTC.
+       There were 17 leap years between 1900 and 1970.
+       Leap seconds started from 1972 so are irrelevant here.
```
akimd: I'm learning things!

djs55:
This has been superseded by running an external server and using `--udp-forwards` to forward to it.

Issue #385, opened 2018-05-13
vpnkit memory leak regression


djs55:
There are reports on #371 that a memory leak was introduced somewhere between 17.12.x and 18.03.y. Unfortunately I've not been able to reproduce it locally (yet).

Here is a table with links to the suspect builds of vpnkit:

| vpnkit PR | macOS build number | Windows build |
|----------|--------|---------|
| #375 | [1175](https://1176-58395340-gh.circle-artifacts.com/0/Users/distiller/vpnkit/vpnkit.tgz) | [1.0.440](https://ci.appveyor.com/api/buildjobs/60uxelglgajqvorl/artifacts/vpnkit.exe) |
| #370 | [1162](https://1162-58395340-gh.circle-artifacts.com/0/Users/distiller/vpnkit/vpnkit.tgz) | [1.0.429](https://ci.appveyor.com/api/buildjobs/qb4nfn9erxq08cc7/artifacts/vpnkit.exe) |
| #369 | [1159](https://1159-58395340-gh.circle-artifacts.com/0/Users/distiller/vpnkit/vpnkit.tgz) | [1.0.423](https://ci.appveyor.com/api/buildjobs/8e8v82biatl3nldh/artifacts/vpnkit.exe) |
| #372 | [1158](https://1158-58395340-gh.circle-artifacts.com/0/Users/distiller/vpnkit/vpnkit.tgz) | [1.0.425](https://ci.appveyor.com/api/buildjobs/4ow1k2uhdsx6gfc5/artifacts/vpnkit.exe) |
| #367 | [1153](https://1153-58395340-gh.circle-artifacts.com/0/Users/distiller/vpnkit/vpnkit.tgz) | [1.0.420](https://ci.appveyor.com/api/buildjobs/75y246jjjd4of99d/artifacts/vpnkit.exe)
| #366 | [1150](https://1150-58395340-gh.circle-artifacts.com/0/Users/distiller/vpnkit/vpnkit.tgz) | [1.0.416](https://ci.appveyor.com/api/buildjobs/dcsi78jaqialwne8/artifacts/vpnkit.exe) |
| #365 | [1147](https://1147-58395340-gh.circle-artifacts.com/0/Users/distiller/vpnkit/vpnkit.tgz) | [1.0.414](https://ci.appveyor.com/api/buildjobs/ydjtkckc8q69gcxd/artifacts/vpnkit.exe) |
[...]

Issue #374, opened 2018-04-12
VPNKIT huge memory consumption, exhausting hard disk bandwitdh when swapping


CdeMills:
Hello,
context: Windows 10 platform,
Version: 18.03.0-ce
API version: 1.37
Go version: go1.9.4
Git commit: 0520e24
Built: Wed Mar 21 23:06:28 2018
OS/Arch: windows/amd64
Experimental: false
Orchestrator: swarm

The docker container is based on fedora27, created as
docker create -t --name fedora --publish 2242:22 --publish 4732:3389 --init --cap-add SYS_ADMIN --cap-add DAC_READ_SEARCH -cap-add=NET_ADMIN --device=/dev/net/tun fedora27 /usr/bin/systemctl

with systemctl coming from https://github.com/gdraheim/docker-systemctl-replacement

The only enabled service is sshd; once the container is started I launch a graphical session with x2go server and a get a full linux environment running on top of Windows10.

After around two hours, the hard disk start doing continuous noise, and vpnkit.exe (in Windows) start consuming all the disk bandwidth. At this time the x2go client becomes unresponsive and disconnects. The only option is to stop the container, kill vpnkit, stop docker. Sometimes the x2go client becomes unresponsive too, requiring a machine reboot.

This occured two times this week.

Is there some way to limit vpnkit memory consumption ?

Regards

Pascal

djs55:
This is the same as https://github.com/moby/vpnkit/issues/371 

There's a candidate fix available. If you'd like to try it before it's officially released there are instructions here: https://github.com/moby/vpnkit/issues/371#issuecomment-390248401

Issue #371, opened 2018-03-25
vpnkit.exe eating my RAM


soar:
This happened after updating to 18.* branch. At this moment I have `18.03.0-ce-rc4-win57 (16511)` build and my `vpnkit.exe` steals gigabytes of RAM in 2-3 hours with 2 containers running.

Like this:

![2018-03-25 18-00-20](https://user-images.githubusercontent.com/30570/37876565-df48c5e2-3056-11e8-8588-7991868e47e0.png)

Or even like this:

![2018-03-24 16-10-41](https://user-images.githubusercontent.com/30570/37876568-e907caf6-3056-11e8-88aa-1c781fbde09f.png)

I think it should never consume about 8 GBs of my RAM.

soar:
Better and better...

![2018-03-25 18-38-54](https://user-images.githubusercontent.com/30570/37876889-ebb4fc56-305b-11e8-8235-edca1e9802ee.png)

bartoncasey:
I have this issue as well. vpnkit.exe claims memory in proportion to the amount of network activity into and out of the docker containers, and never releases it.

bartoncasey:
1 day later it's sitting at 1.2G, from light traffic.

laarmen:
We are several here at work having this exact issue.

djs55:
Thanks for your reports.

In order to make progress with this issue I need some reproduction steps. Could you provide a `docker-compose.yml` (or similar) and instructions to reproduce the problem?

laarmen:
After trying a bit, it seems opening an HTTPS connection to a server on our internal network triggers the bug. The same doesn't apply to external, public servers (i.e. docker.com) nor other Docker instances.

djs55:
@laarmen thanks for the update. Could you trigger the bug and then upload a diagnostic report? I'd like to take a look at the logs.

laarmen:
See https://github.com/laarmen/VpnKitPoC for the code. How can I do the diagnostic report thing?

djs55:
On Windows there should be a whale-shaped icon in the system tray. After
right clicking on it there should be a menu item called something like
"Diagnose and Feedback". Clicking on this should take you to a dialog where
diagnostics are uploaded and assigned a unique id. If you quote the id in
the ticket then I can download the logs and take a look.

(Sorry I couldn't give more precise instructions but I don't have a Windows
machine to hand)

laarmen:
[...]

PR #389, opened 2018-05-18
Revert TCP keepalives to reduce memory usage

djs55:
The TCP keep-alive code from https://github.com/mirage/mirage-tcpip/pull/338 has a flaw -- when a connection has been closed by the remote then when the callback fires and sends an `ACK` it will provoke a `RST` but this is ignored and the timer reset. The receipt of a `RST` should cancel the keep-alive timer since the connection is definitely not "alive". Since it ignores the `RST`, the keep-alive thread holds on to the per-flow state until the final timer expires.

Each TCP connection has send and receive buffers containing data not sent and not read respectively. If lots of data is sent to a peer which isn't calling `read` then the data (up to one full window) will be cached in the receive buffers. If the connection is then closed but the keep-alive thread still has a reference to the state, then the send and receive buffers will also be retained.

Therefore we can cause `vpnkit` to allocate more and more memory by creating lots of TCP connections, sending lots of data (but not reading it) and then closing the connection. The accumulation of per-connection state will look like a memory leak.

For a concrete example consider a process in a container attempting to upload a file by `HTTP POST` to a server on the host, where the server on the host is buggy and closes the connection "immediately". Even if the user-space application on the host calls `close` immediately, there will still be time for the client to have sent the initial portion of the upload. This data will be cached by vpnkit until the connection is closed at the TCP-level. This state cannot be GCed until the keep alive-timer has dropped its reference to it and therefore "leaks" (temporarily)

This can be tested with a simple program like the following run inside a container:

```go
package main

// Run in a container to send large amounts of data to the HTTP proxy.
// This will fill the receive buffers in the remote.

import (
  "log"
  "net"
  "fmt"
  "time"
)

var zeroes = make([]byte, 1048576)

func main() {
  for {
    one()
    fmt.Printf(".\n")
  }
}

func one() {
  // connect to this socket
  conn, err := net.Dial("tcp", "docker.for.mac.gateway.internal:3128")
  if err != nil {
    log.Fatal(err)
  }
  defer conn.Close()

  req := []string{
    "POST / HTTP/1.1",
    "Host: localhost:80",
    "User-Agent: curl/7.54.0",
    "Accept: */*",
    "Content-length: 1048576",
    "Connection: close",
  }
  for _, line := range req {
    conn.Write([]byte(line))
    conn.Write([]byte("\r\n"))
  }
  conn.Write([]byte("\r\n"))

  // write to socket with a deadline
  conn.SetDeadline(time.Now().Add(100 * time.Millisecond))
  conn.Write(zeroes)
}
```
talking to a program like this on the host:
```go
package main

// Receive HTTP connections on port 80, read some data and then
// close them prematurely. The TCP stack will still have un-read
// data in its buffers.

import (
    "bufio"
    "fmt"
    "log"
    "net"
    "os"
    "time"
)

const (
    CONN_HOST = "localhost"
    CONN_PORT = "80"
    CONN_TYPE = "tcp"
)

func main() {
    // Listen for incoming connections.
    l, err := net.Listen(CONN_TYPE, CONN_HOST+":"+CONN_PORT)
    if err != nil {
        fmt.Println("Error listening:", err.Error())
        os.Exit(1)
    }
    // Close the listener when the application closes.
    defer l.Close()
    fmt.Println("Listening on " + CONN_HOST + ":" + CONN_PORT)
    for {
        // Listen for an incoming connection.
        conn, err := l.Accept()
        if err != nil {
            fmt.Println("Error accepting: ", err.Error())
            os.Exit(1)
        }
        // Handle connections in a new goroutine.
        go handleRequest(conn)
    }
}

// Handles incoming requests.
func handleRequest(conn net.Conn) {
  defer conn.Close()
  fmt.Printf("Accepted connection\n")
  buf := make([]byte, 1024)
  conn.Read(buf)

  if false {
    b := bufio.NewReader(conn)
    for {
      line, err := b.ReadString('\n')
      if err != nil {
        log.Fatal(err)
      }
      if line == "\r\n" {
          break
      }
      fmt.Printf("%s\n", line)
    }
  }
  fmt.Printf("responding\n")
  conn.Write([]byte("HTTP/1.0 100 Continue\r\nConnection: close\r\n\r\n"))
  fmt.Printf("closing\n")
  time.Sleep(100 * time.Millisecond)
}
```

Without this PR, this code causes the memory footprint (as measured by Activity Monitor) to rise > 1 GiB within about 1 minute. With this PR, the memory footprint hovers around 300-400 MiB over 1 hour.

Related to #385 #374 #371
Reverts #359


djs55:
There are several reports of success on the main issue so I think this is good to merge.

PR #410, opened 2018-09-04
vpnkit-tap-vsockd: update logging, add a post-up script

djs55:
Previously we used a mixture of `DEBUG` `INFO` macros and calls to `syslog`. We had a `verbose` flag in the code but this wasn't hooked up to the command-line. This PR makes the code use logging macros everywhere and hooks up the `verbose` flag, removing `syslog` because there is no `syslog` in LinuxKit VMs.

Since a user of this will want to configure the interface once it has been brought up, this PR also adds a command-line `--post-up-script` which will be called as soon as the interface is created. The script can be used to run DHCP or otherwise configure the IP address of the interface. To make use of this easier, the final image has been upgraded to an `alpine` with the usual scripting tools.

### DIFF
--- a/c/vpnkit-tap-vsockd/Dockerfile
+++ b/c/vpnkit-tap-vsockd/Dockerfile
@@ -4,6 +4,8 @@ RUN apk add --no-cache go musl-dev build-base linux-headers
 COPY . /build
 RUN make -C /build sbin/vpnkit-tap-vsockd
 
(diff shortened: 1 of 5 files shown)


djs55:
CircleCI and appveyor don't compile this code and are broken for different reasons (CircleCI infrastructure and a possible Cygwin problem with appveyor).

PR #411, opened 2018-09-04
vpnkit-tap-vsockd, vpnkit-9pmount-vsock: try AF_VSOCK first, fall back to AF_HVSOCK

djs55:
The `AF_HVSOCK` patches were never merged in upstream Linux, instead Hyper-V support was added to `AF_VSOCK` instead. One consequence is the `AF_VSOCK` "port" is mapped into a special space of Hyper-V service GUIDs. This patch modifies `vpnkit-tap-vsockd` and `vpnkit-9pmount-vsock` to first try `AF_VSOCK` and then fall back to `AF_HVSOCK`.

This patch changes the command-lines from
```
      --serviceid <guid>
```
 to
```
      --vsock <port>
```

### DIFF
--- a/c/vpnkit-9pmount-vsock/9pmount-vsock.c
+++ b/c/vpnkit-9pmount-vsock/9pmount-vsock.c
@@ -2,7 +2,6 @@
 #include <stdlib.h>
 #include <string.h>
 #include <getopt.h>
-#include <syslog.h>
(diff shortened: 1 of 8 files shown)


djs55:
Only the travis build is relevant, since this is Linux rather than Mac (CircleCI) or Windows (appveyor)

PR #412, opened 2018-09-04
vpnkit-{tap,9pmount}-vsock: log to stderr

djs55:
Previously we would log to stdout.


djs55:
The appveyor CI doesn't test the Linux container code.

PR #418, opened 2018-09-10
Remove the "shutdown" protocol used for forwarding ports over Hyper-V sockets

djs55:
Previously we used a custom protocol on top of Hyper-V sockets because they didn't support `shutdown`. With modern Windows and modern Linux the `shutdown` works (at least when the connections are initiated from the host) and therefore the Go `linuxkit/virtsock` library has dropped the custom protocol. Since we're talking to `vpnkit-forwarder` using `linuxkit/virtsock` we must drop the custom protocol too.

This uses an experimental version of `mirage/ocaml-hvsock` with support for syscall-based `shutdown`.


djs55:
The hvsock API has changed. I'll close this PR and open a fresh one.

PR #424, opened 2018-10-06
Remove the old shutdown protocol

djs55:
The old protocol was needed to work around Hyper-V socket bugs, where data would be lost after a `shutdown`. This is no longer needed because the new multiplexing protocol doesn't call `shutdown`, since it multiplexes all port forwarding connections over a single connection.

This supercedes #418

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+### v0.4.0 (unreleased)
+* allow longer paths when forwarding Unix domain sockets
+
 ### v0.3.0 (2019-02-06)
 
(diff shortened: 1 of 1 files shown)


djs55:
I've rebased and temporarily removed the changes on the Go side.

So now
- when `vpnkit` connects to `vpnkit-forwarder` it will not use the old `shutdown` protocol
- when `vpnkit-forwarder` listens on `AF_VSOCK` it will not use the old `shutdown` protocol
- when `vpnkit-forwarder` listens on `AF_HVSOCK` it *will* use the old `shutdown` protocol <- this will not work

This PR is only useful for testing against newer kernels.

PR #426, opened 2018-10-12
vmnet: move vmnet to its own package

tiborvass:
This allows library users to import vmnet without datakit.

Signed-off-by: Tibor Vass <tibor@docker.com>

### DIFF
--- a/go/pkg/vmnet/vmnet.go
+++ b/go/pkg/vmnet/vmnet.go
@@ -1,4 +1,4 @@
-package vpnkit
+package vmnet
 
 import (
 	"bytes"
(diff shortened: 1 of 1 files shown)


samoht:
LGTM but I am just curious about the context in which this change is needed as I though vmnetd was already independant of datakit.

tiborvass:
@samoht it's used in https://github.com/rootless-containers/rootlesskit and we discovered the unnecessary datakit dependency. Right now, the other files from the vpnkit package depend on datakit.

tiborvass:
@samoht and we use rootlesskit in buildkit :)

djs55:
Also LGTM!

PR #447, opened 2019-01-05
Update to opam2

djs55:
The Linux `Dockerfile` already works with `opam2`, which is a good start.

- remove the `opam` 1.2 pin from CircleCI
- update the download links for AppVeyor

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,8 +15,7 @@ jobs:
     steps:
       - checkout
       - run: brew uninstall python # remove files in /usr/local/bin
-      - run: brew install wget pkg-config dylibbundler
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
Any update?

yamt:
@djs55 are you still working on this?

djs55:
I've not got the time at the moment :( Feel free to have a play with it if you want!

djs55:
CircleCI is working with opam2 but Appveyor is still broken due to a cygwin problem. I think we need to upgrade many of the dependencies and try to remove the remaining (build-time) uses of cygwin. Maybe we can switch to a vendored model via [opam monorepo](https://github.com/ocamllabs/opam-monorepo) to simplify the build even further (perhaps we can have `git clone; dune build`)

PR #450, opened 2019-01-25
Add Kubernetes controller used by Docker Desktop

guillaumerose:
This PR also contains:
* Add Kubernetes controller used by Docker Desktop. For the moment, just as a package, that can be used on the host (not in the VM yet).
* Move from vndr to go dep: dep was already used by the binary `kube-vpnkit-forwarder`. It's too complicated to vendor Kubernetes with vndr.
* Fix one go unit test failure

One question, I have: vendoring of linuxkit/virtualsock seems to be really old. Can we update it ?


justincormack:
I beleive the vsock vendoring is old as you are still running patched 4.9 kernels which is before the interface changes for hvsock.

djs55:
Yes unfortunately we can't update the vsock version until we've bumped the kernel version (which requires some more bugfixing).

guillaumerose:
Can I use the tag `end_10586_support` or `10586_fixes` in virtsock ?

djs55:
The changes in `10586_fixes` look safe enough (and some of them could be useful)

guillaumerose:
Thanks!

Le ven. 25 janv. 2019 17:29, David Scott <notifications@github.com> a
écrit :

PR #452, opened 2019-01-30
test: fix TestUDPWriteError

djs55:
The client is supposed to send UDP to the "frontend" of the IP proxy
to check the proxy isn't stopped when it tries to forward to a "backend"
which doesn't exist yet.


djs55:
FYI @guillaumerose I fixed that broken unit test

PR #462, opened 2019-02-13
Restore the vpnkit-userland-proxy

djs55:
`dockerd` allows the userland proxy to be overridden with a custom binary. In Docker Desktop we use the LinuxKit `pkg/vpnkit-expose-port` which is now different from our `vpnkit-expose-port`. The `vpnkit-expose-port` in this repo is an example of how to use the `vpnkit` Go package to manipulate ports; it does not know how to manipulate a mounted 9P filesystem like the `pkg/vpnkit-expose-port` does.

This PR
- restores the old `pkg/vpnkit-expose-port` as `cmd/vpnkit-userland-proxy` (to use the same terminology as Docker)
- sets the `-no-local-ip` parameter to `true` to avoid port clashes in the VM


djs55:
This PR only involves Go code, so it's safe to merge without waiting for appveyor.

PR #463, opened 2019-02-13
Remove `make test` from `vpnkit-userland-proxy`

djs55:
The `vpnkit-userland-proxy` doesn't use any of the common libraries.

This speeds up the build.


djs55:
Travis has built successfully.

djs55:
Travis has built successfully.

PR #464, opened 2019-02-13
go: improve build of vpnkit-userspace-proxy

djs55:
- simplify the `Dockerfile` by removing `make` and following the LinuxKit standard
- excluding the Dockerfiles from the build context to avoid unnecessary rebuilds

### DIFF
--- a/go/.dockerignore
+++ b/go/.dockerignore
@@ -1,0 +1,1 @@
+Dockerfile*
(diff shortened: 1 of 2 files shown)


djs55:
Travis is happy, let's see if the docker hub auto builder is similarly happy.

PR #465, opened 2019-02-13
go: use COPY instead of ADD in Dockerfile

djs55:
COPY is less powerful than ADD so prefer COPY in this case.


djs55:
Thanks!

PR #466, opened 2019-02-14
When calling OnUpdate, Kubernetes controller now closes removed ports

guillaumerose:
The current implementation of the Kubernetes controller doesn't handle update properly. 

For instance, if I deploy an nginx and expose it:
```
kubectl run --image nginx nginx
kubectl expose deploy nginx --port 80 --type=NodePort
```
Then after change the port using `kubectl edit svc nginx`.
Two ports are now opened. The original port should be close.


djs55:
Looks good to me!

PR #473, opened 2019-04-03
fix regression in "vmnet: remove dhcp pcap" (#472)

AkihiroSuda:
Signed-off-by: Akihiro Suda <suda.akihiro@lab.ntt.co.jp>

sorry #472 caused regression

@djs55


djs55:
I think I need to write some unit tests for this!

PR #475, opened 2019-05-01
vpnkit-userland-proxy: bind ports in the VM on a best-effort basis

djs55:
When Docker attempts to expose a port on the host with
```
  docker run -p <external>:<internal>
```
We have to decide whether to also bind the same port in the VM. In the case of addresses like `0.0.0.0:80` or `127.0.0.1:80` this can make sense, and it will allow running a Docker registry in a container and pushing directly to it, see [docker/for-mac#3611]

However it also opens us up to accidental port clashes between the user's ports and any that we have allocated internally. Recently this happened when a compose on kubernetes container was run with `--net=host` and bound port 8080, see [docker/compose-on-kubernetes#70].

This patch adds support for "best-effort" binding in the VM and makes this the default. This should re-enable the registry use-case while making us robust to the compose on kubernetes problem. The only downside is that if there is a port clash in the VM, the user won't be notified.

### DIFF
--- a/go/cmd/vpnkit-userland-proxy/proxy.go
+++ b/go/cmd/vpnkit-userland-proxy/proxy.go
@@ -43,9 +43,17 @@ var vSockUDPPortOffset = 0x20000
 
 // From docker/libnetwork/portmapper/proxy.go:
 
+type localBind int
+
+const (
+	bestEffortLocalBind = localBind(0)
+	alwaysLocalBind     = localBind(1)
+	neverLocalBind      = localBind(2)
+)
+
 // parseHostContainerAddrs parses the flags passed on reexec to create the TCP or UDP
 // net.Addrs to map the host and container ports
-func parseHostContainerAddrs() (host net.Addr, port int, container net.Addr, localIP bool) {
+func parseHostContainerAddrs() (host net.Addr, port int, container net.Addr, bind localBind) {
 	var (
 		proto         = flag.String("proto", "tcp", "proxy protocol")
 		hostIP        = flag.String("host-ip", "", "host ip")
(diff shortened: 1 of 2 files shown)


go/cmd/vpnkit-userland-proxy/proxy.go
```diff
@@ -43,17 +43,25 @@ var vSockUDPPortOffset = 0x20000
...
+func parseHostContainerAddrs() (host net.Addr, port int, container net.Addr, bind localBind) {
 	var (
 		proto         = flag.String("proto", "tcp", "proxy protocol")
 		hostIP        = flag.String("host-ip", "", "host ip")
 		hostPort      = flag.Int("host-port", -1, "host port")
 		containerIP   = flag.String("container-ip", "", "container ip")
 		containerPort = flag.Int("container-port", -1, "container port")
 		interactive   = flag.Bool("i", false, "print success/failure to stdout/stderr")
-		noLocalIP     = flag.Bool("no-local-ip", true, "bind only on the Host, not in the VM")
+		local         = flag.String("no-local-ip", string(bestEffortLocalBind), "bind only on the Host, not in the VM")
```
ijc: The flag `--no-local-ip` doesn't really make sense now I think.

djs55: That's true. There are a couple of scripts floating around that have `-no-local-ip` (without an argument) which I'd like to keep working for a little while. I've pushed an update with
- `-no-local-ip` removed from the help text
- a new `-local-bind (best effort | always | never)` argument
- a special-case search of the arguments for `-no-local-ip` as a flag which sets the default to `never` as a temporary workaround for the existing scripts

I propose merging something like this, then separately modifying the scripts and when those have been released, coming back here and removing the workaround. WDYT?

ijc: SGTM.

djs55:
Thanks!
I'll merge ignoring the appveyor CI failure since it's is irrelevant (although annoying that I haven't had the time to fix it yet)

PR #478, opened 2019-07-01
Update to new virtsock API

djs55:
The upstream linuxkit/virtsock API is different from that in the `10586_fixes` branch that we're using.

This PR uses a backport of the new API onto the old branch. This is useful since we're still using the old kernel in Desktop.


djs55:
Travis (CI for the Go code) is green, so I think we can merge this.

PR #480, opened 2019-07-19
Fix Dockerfile

### LINKED ISSUE #455: Dockerfile fails to build
Building the `Dockerfile` fails me. Both using buildkit and legacy build.

screen shot 2019-02-04 at 11 20 20

```
$ DOCKER_BUILDKIT=1 docker build -t vpnkit .
#12 313.3 <><> jbuilder.transition installed successfully <><><><><><><><><><><><><><><><>
#12 313.3 => Jbuilder has been renamed and the jbuilder package is now a transition package. Use the dune package instead.
#12 313.3 # Run eval $(opam env) to update the current shell environment
#12 313.4
#12 313.4 The former state can be restored with:
#12 313.4     opam switch import "/home/opam/.opam/4.07/.opam-switch/backup/state-20190204100157.export"
```

djs55:
Fixes #455

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ocaml/opam2:alpine as build
+FROM ocaml/opam2:alpine-3.10-ocaml-4.07 as build
 ADD . /home/opam/src
 # Latest packages plus some overrides are in this repo:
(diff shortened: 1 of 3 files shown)


djs55:
I don't think we have a CI step which specifically builds the Dockerfile :(

PR #481, opened 2019-07-23
Go: improve the logging

djs55:
- add `logrus`
- use `Errorf` where appropriate
- don't log "expected" errors around closing
- attempt to hide transient Unix domain socket errors by retrying for up to 120s, assuming that these are forwards to internal services (Unix domain socket forwarding is not used by Docker)


djs55:
Ignoring the OCaml CI, the Travis CI is green for Go. This also passed the Docker Desktop end-to-end test suite which is quite reassuring.

PR #484, opened 2019-08-05
Use unix sockets in vpnkit-tap-vsockd

simonferquel:
This changes the vpnkit-tap-vsockd to avoid using HV-Sockets directly but using a unix socket


simonferquel:
@djs55 Not sure about the CI errors here. Is it only transient errors or did I break something ?

djs55:
I think the CI errors are problems with the OCaml code -- since this touches only the `vpnkit-tap-vsockd` process I think we can ignore them. (I know the appveyor CI is broken because of a Cygwin problem. I'm not sure why CircleCI is currently broken)

Unfortunately there are no tests for `vpnkit-tap-vsockd`. Perhaps the next step is to build this into an image, push it to hub and then try to test it end-to-end?

simonferquel:
Do you want me to do this as part of this PR?

simonferquel:
Btw, is there a linux build of vpnkit.exe somewhere ? IIUC the Makefile, it does only build for Mac and Windows.

What I wanted to do, is to run vpnkit.exe on the host, run a container with none network, bind mount the vpnkit socket and run vpnkit-tap-vsockd to bring connectivity, and validate this with docker exec `wget http://www.google.com` within this container.

simonferquel:
Added the e2e test, and ran it in travis :)

PR #485, opened 2019-08-05
go: fix binding of privileged UDP ports via vmnetd

djs55:
Previously in 32ede13e84c08bdc76ac8ea17d5e84863b89637f we added a fix for TCP where we didn't set the socket to non-blocking mode. This PR replicates the fix for UDP and adds a unit test.

Related to docker/for-mac#3775

Furthermore this PR only uses the `vmnetd` code path if the bind actually fails with permission denied. In particular on Mojave, some binds will succeed (e.g. `0.0.0.0:80`) without being root.


djs55:
Ignoring the OCaml broken CIs, travis is green for the Go code. This also passes the Docker Desktop end-to-end tests for extra safety.

PR #487, opened 2019-09-06
mux: perform an initial handshake to check the data connection

djs55:
Previously if we connected to the wrong service we would output a confusing set of error messages.

Previously if the connection returned EOF (e.g. because it was forwarded by something like `docker run -p`) we would also output a confusing set of error messages.

This patch makes the constructor `NewMultiplexer` return an error, which is the result of a handshake exchange. This means that `NewMultiplexer` is now a blocking operation.

The handshake contains the magic string
```
  https://github.com/moby/vpnkit multiplexer protocol\n
```
so hopefully it will be obvious to other people what this protocol is.

For future use the handshake includes a variable length payload which could contain feature flags in the future.

### DIFF
--- a/go/pkg/libproxy/handshake.go
+++ b/go/pkg/libproxy/handshake.go
@@ -1,0 +1,50 @@
+package libproxy
+
+import (
+	"encoding/binary"
+	"fmt"
+	"io"
+)
+
+type handshake struct {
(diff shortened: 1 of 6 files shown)


djs55:
Travis (the CI which builds and tests the Go code) is now happy, so merging

PR #494, opened 2019-11-08
Remove `is being closed.` logs in proxystream

ebriney:
They happen every time the client named pipe do a close and it don't hide any errors

(it's quite similar to `is not connected` filtering)


djs55:
The CircleCI and Appveyor CIs are both irrelevant to this PR: it should be tested by travis. I think it's safe to merge.

PR #495, opened 2019-11-16
Fix cross-compilation

djs55:
Previously we had to reference both `hvsock` and `vsock` from Linux code because we supported old 4.9 Linux kernels with the non-upstream `AF_HVSOCK` patches applied. The only user of these patches was Docker Desktop but it has now upgraded to vanilla upstream `AF_VSOCK` in 4.19.

Previously cross-compilation was broken, e.g.:
```
$ GOOS=linux go build
../../../vendor/github.com/linuxkit/virtsock/pkg/hvsock/hvsock.go:158:10: undefined: hvsockListener
../../../vendor/github.com/linuxkit/virtsock/pkg/hvsock/hvsock.go:173:10: undefined: hvsockListener
../../../vendor/github.com/linuxkit/virtsock/pkg/hvsock/hvsock.go:178:10: undefined: hvsockListener
../../../vendor/github.com/linuxkit/virtsock/pkg/hvsock/hvsock.go:205:2: undefined: hvsockConn
```

This PR drops the old `AF_HVSOCK` code and untangles the `hvsock` and `vsock` imports to fix cross-compilation.


djs55:
Travis CI fixed, CircleCI and Appveyor are known to be broken already.

PR #500, opened 2020-02-24
Remove datakit implementation of go client

guillaumerose:
Some cleanup in Go packages. I would like to reduce the number of dependencies of vpnkit package (esp. remove the link to labstack echo).


djs55:
I've force-pushed a rebase to master now that travis is fixed.

djs55:
Travis is green! Merging

PR #503, opened 2020-03-05
Isolate vpnkit http server in a separate package

guillaumerose:
It’s required in order to avoid to vendor labstack/echo when using the vpnkit client.


guillaumerose:
Docker Desktop and Linuxkit CI are happy with this change. Merging it.

PR #504, opened 2020-03-05
Upgrade linuxkit/virtsock vendoring

guillaumerose:
Docker Desktop vendors vpnkit master + virtsock master and it works like a charm. Let's synchronize it here also.


djs55:
Travis is green and since it's the only CI that builds the Go code I think we're good to merge.

PR #506, opened 2020-07-16
transport: work around Unix socket path lengths on Darwin

djs55:
On Darwin a Unix socket path can only be 104 bytes long. We use long paths when talking to hyperkit
(`~/Library/Containers/com.docker.docker/...`) so we often hit this limit if the user has a long username.

A common workaround is to `chdir` into the `~/Library/Containers/com.docker.docker` folder (as would happen naturally if running in a macOS sandbox) and then use relative paths.

This patch checks the path lengths and constructs a relative path if necessary.


djs55:
Sadly yes appveyor and circleci are broken :( The good news is that travis tests the Go code 😅

PR #507, opened 2020-07-24
transport: fix off-by-one in the max socket path

djs55:
On Mac the max socket length is 104 which includes a NULL on the end. A Go path of length 104 will fail. The effective max length is 103.

This patch includes a test case which checks that, when the bind fails, the length is > maxUnixSocketPathLen.

Signed-off-by: <David Scott dave.scott@docker.com>

### DIFF
--- a/go/Gopkg.lock
+++ b/go/Gopkg.lock
@@ -413,9 +413,12 @@
   version = "v1.4.0"
 
 [[projects]]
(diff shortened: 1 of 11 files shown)


thaJeztah:
Do we need a "lol 104 chars ETOOMANY"? https://github.com/moby/moby/pull/13408

djs55:
@thaJeztah hah yes, just when you thought 108 chars was bad, now we have 104!

@guillaumerose thanks!

PR #510, opened 2020-12-02
go: remove hardcoded GOARCH to support multiarch builds

djs55:


guillaumerose:
I really don't understand why you do that ahah :angel:

Perhaps, `GOARCH?=amd64`. Otherwise, lgtm.

djs55:
Yes no particular reason why I'm suddenly interested in multiarch. None at all!

PR #512, opened 2021-01-22
Handle random port correctly: report assigned port

simonferquel:
This makes sure forwarded UDP and TCP services on random ports report back their randomly assigned ports.


djs55 (approved):
LGTM

djs55:
Travis is green and it tests the Go code so I think we're good.

PR #514, opened 2021-01-22
go: switch to go.mod, Mac build fixes

djs55:
Although the go code is tested by CI on Linux, the Mac CI is currently broken and doesn't pick up failures in Mac-specific parts.

PR #515, opened 2021-01-22
go: use narrower UDPListener interface

djs55:
This is smaller and easier to get than a full *net.UDPConn.

Split from #514 in case I mess up the `go.mod` transition at the same time.

Unfortunately the CI builds the Go code under Linux but the Mac CI is broken, allowing this kind of breakage to slip through.


simonferquel:
PR passed on Docker Desktop side. Good to merge!

PR #520, opened 2021-04-15
circleci: remove `brew update` which takes 5 minutes

djs55:
Let's see if we can speed up the Mac CI.


djs55:
It looks like the brew setup in the CircleCI workers is out-of-date and needs updated :(

PR #523, opened 2021-04-17
Improve / simplify logging

djs55:
Previously we supported 3 log "destinations":
1. `quiet`: stderr
2. `asl`: the Apple System Log
3. `eventlog`: the Windows event log

This PR removes the `asl` and `eventlog` systems to simplify the build dependencies. The `eventlog` mode was never used anywhere and I'm not sure if it's implemented correctly. The `asl` mode is being used in Docker Desktop but this isn't a good idea: the system logs are difficult to filter and full of spam. It's much easier to simply log to stderr and rely on another tool to send the logs to the correct destination (e.g. `docker logs`)

At the same time the output format has been changed to resemble other Go programs for consistency, e.g.:

```
time="2021-04-17T10:56:51Z" level=info msg="usernet: Increasing preemptive thread pool size to 1024 threads"
```

Plus a few error logs have been downgraded to info where they are normal e.g. `/etc/resolv.conf` doesn't exist on Windows; receiving `EOF` from a disconnecting client is normal.

### DIFF
--- a/src/bin/bind.ml
+++ b/src/bin/bind.ml
@@ -16,9 +16,9 @@ module Make(Socket: Sig.SOCKETS) = struct
 
   module Channel = Mirage_channel_lwt.Make(Socket.Stream.Unix)
 
-  let err_eof = Lwt_result.fail (`Msg "error: got EOF")
-  let err_read e = failf "error while reading: %a" Channel.pp_error e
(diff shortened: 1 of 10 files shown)


djs55:
CircleCI is green!

PR #525, opened 2021-04-17
jbuilder is deprecated, so use dune (the new name for jbuilder)

djs55:


djs55:
CircleCI is green!

PR #527, opened 2021-04-18
Base idle time calculations on a CPU counter, not a wall clock

djs55:
Previously we used the wall clock via `Unix.gettimeofday()` to compute the idle time for "endpoints" (remote destinations). When a laptop wakes from sleep, the wall clock jumps forward and suddenly all the endpoints are idle and the connections are dropped.

This patch switches to a monotonic clock which is driven by a CPU counter. The counter is paused while the computer sleeps and therefore does not jump forwards on wake.

This patch also adds a log message whenever an "endpoint" is deleted, to make debugging problems easier.

### DIFF
--- a/src/hostnet/slirp.ml
+++ b/src/hostnet/slirp.ml
@@ -192,16 +192,24 @@ struct
       (** An established flow *)
 
       type t = {
+        clock: Clock.t;
         id: Stack_tcp_wire.t;
         mutable socket: Host.Sockets.Stream.Tcp.flow option;
-        mutable last_active_time: float;
+        mutable last_active_time_ns: int64;
       }
 
+      (* Consider the case when a computer is sleep for several hours and then powered on.
+         If we use the wall clock: we would conclude that all flows have been idle for hours
+         and possibly terminate them.
+         If we use a monotonic clock driven from a CPU counter: the clock will be paused while the
+         computer is asleep so we will conclude the flows are still active. *)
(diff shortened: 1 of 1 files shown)


src/hostnet/slirp.ml
```diff
@@ -239,14 +249,16 @@ struct
...
-      mutable last_active_time: float;
+      mutable last_active_time_ns: int64;
       (* Used to shutdown connections when the endpoint is removed from the switch. *)
       mutable established:      Tcp.Id.Set.t;
     }
     (** A generic TCP/IP endpoint *)
 
     let touch t =
-      t.last_active_time <- Unix.gettimeofday ()
+      t.last_active_time_ns <- Clock.elapsed_ns t.clock
```
djs55, line 259: `Clock.elapsed_ns` calls into https://github.com/mirage/mirage-clock/blob/987d197802f80778b7b3daa642caee0c27a1ce43/unix/mclock.ml#L19 which is https://github.com/mirage/mirage-clock/blob/987d197802f80778b7b3daa642caee0c27a1ce43/unix/clock_stubs.c#L67 

According to https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time

> Returns current value of a clock that increments monotonically in tick units (starting at an arbitrary point), this clock does not increment while the system is asleep.

StefanScherer (approved):
I've tested a Docker Desktop build on a M1 machine leaving unattended over night and Docker is still working and docker commands from the command line still work as expected. The machine went to sleep mode (I have disabled the "Prevent from sleeping" system setting in macOS).

PR #528, opened 2021-04-18
CircleCI: restore the COMMIT and OSS-LICENSES artifacts

djs55:
These were lost when the CI was fixed. Minor tweaks are required to update the scripts to support `opam` version 2.


djs55:
CircleCI is green and the artifacts are back!

PR #529, opened 2021-04-19
ounit: correct download URL

djs55:


avsm:
Confirmed as as same as opam upstream urls for these.

PR #530, opened 2021-04-20
Update OCaml to 4.12

djs55:
This one has Apple Silicon support!

A few package pins are necessary (temporarily)


djs55:
CircleCI is green and the OCaml subset of travis is green: looks like the dependency upgrade is ok.

PR #531, opened 2021-04-21
Exposing a port should be idempotent and fix an ephemeral port leak

djs55:
Previously attempting to expose an existing port would return an error, which becomes an HTTP 500 internal server error.

This patch makes exposing a port idempotent, just like un-exposing a port already is.

Previously there was a port leak if the user requested any port (`0`) because we would store `0` in our map, and then look up the real port in `Unexpose` and fail to find it.

To make the API clearer, we switch to `PUT` and `DELETE` (instead of `POST`) but for backwards compatibility retain the `POST` handlers.

Minor additional updates:
- switch from `*sync.Mutex` to `sync.Mutex` where we are already using a pointer receiver, since the `nil` value is a valid mutex
- run the unit tests with the race detector enabled
- avoid logging as errors 3 expected conditions (all variations of an `EOF` on shutdown)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ ENV OPAMJOBS=8
 
 RUN opam install --deps-only vpnkit -y
 
-WORKDIR /home/opam/src
+WORKDIR /home/opam/vpnkit
 RUN opam exec -- sudo dune build --profile release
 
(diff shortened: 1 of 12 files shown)


ebriney (approved):
LGTM

PR #532, opened 2021-04-21
Fix and speed up CircleCI

djs55:


djs55:
CircleCI is green and faster!

PR #536, opened 2021-05-03
Update base image and cleanup Dockerfile for kube-vpnkit-forwarder

aiordache:
Update `Dockerfile.kube-forwarder` to use golang:1.16-alpine as parent image.


djs55 (approved):
LGTM

PR #537, opened 2021-05-06
Set O_CLOEXEC on received file descriptors

djs55:
By default Go sets `O_CLOEXEC` on file descriptors it creates. This ensures that they aren't accidentally inherited over fork+exec. 

When binding a port on the host we first try `net.Listen`. On modern macOS this works for
- unprivileged ports (> 1024)
- privileged ports on `INADDR_ANY` (all IPs)

Unfortunately it doesn't work for privileged ports on specific IPs, such as `127.0.0.1:80`. For these cases we contact a helper process which runs as root, which binds the port and then uses `sendmsg` to transmit it to us. Unfortunately when we `recvmsg` a file descriptor, the resulting descriptor does not have `O_CLOEXEC` set, so it is captured by future subprocesses. Even if the parent process calls `close()`, the port is still in use by the subprocesses and can't be re-bound. This caused the failure seen in https://github.com/docker/for-mac/issues/5649

The solution is to ensure all file descriptors have `O_CLOEXEC` set. On Linux there's a `sendmsg` flag to set the flag atomically on received fds, but this is missing on Darwin. The best we can do is to call `fcntl` to set the flag ourselves after receiving the fd. Unfortunately this means there's a small race window where a fork would inherit the fd, but this shouldn't happen very often in practice.

Go will call `fcntl` for us if we "adopt" the file descriptor using `os.NewFile` and then use `net.FileListener` and `net.FilePacketConn`. This allows us to remove a lot of complicated wrapping code and allows us to remove a workaround for `Close` hanging when blocked in `Accept`, caused by the state of the file descriptor not being correct. This required some minor changes to types (`net.Listener` rather than `*net.TCPListener`) but it's worth it.

It's possible to work around this problem in other ways in the client code, for example:

1. always speculatively closing file descriptors after `fork` but before `exec`. This can be expensive with high `ulimit` values, if there is no API to enumerate open fds. I'm not sure it can be done easily in Go where the `exec.Command` API is high-level.
2. ban running long-lived subprocesses in your program. Restructure the code so that the long-lived subprocesses are managed by a separate process which doesn't bind ports.

### DIFF
--- a/go/pkg/vpnkit/forward/vmnet_darwin.go
+++ b/go/pkg/vpnkit/forward/vmnet_darwin.go
@@ -17,38 +17,13 @@ import (
 )
 
 func listenTCPVmnet(IP net.IP, Port uint16) (net.Listener, error) {
-	// I don't think it's possible to make a net.TCPListener from a raw file descriptor
-	// so we use a hack: we create a net.TCPListener listening on a random port and then
-	// use the `SyscallConn` low-level interface to replace the file descriptor with a
-	// clone of the one which is listening on the privileged port.
-	l, err := net.ListenTCP("tcp", &net.TCPAddr{
-		IP:   IP,
-		Port: 0,
-	})
-	if err != nil {
-		// IP address invalid? Fail early
-		return nil, err
-	}
-	raw, err := l.SyscallConn()
-	if err != nil {
-		_ = l.Close()
-		return nil, err
-	}
 	newFD, err := listenVmnet(IP, Port, true)
 	if err != nil {
-		_ = l.Close()
 		return nil, err
 	}
-	// Unfortunately setting non-blocking mode seems to break I/O in interesting ways,
-	// we we have to leave it in blocking mode. Unfortunately this makes `Close` more complicated,
-	// see below.
-	if err := switchFDs(raw, newFD); err != nil {
-		_ = l.Close()
-		_ = syscall.Close(int(newFD))
-		return nil, err
-	}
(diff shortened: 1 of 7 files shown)


go/pkg/vpnkit/forward/vmnet_darwin.go
```diff
@@ -142,16 +142,38 @@ func closeUDPVmnet(IP net.IP, Port uint16, l libproxy.UDPListener) error {
...
 func switchFDs(raw syscall.RawConn, newFD uintptr) error {
-	var dupErr error
+	var controlErr error
 	err := raw.Control(func(fd uintptr) {
-		dupErr = syscall.Dup2(int(newFD), int(fd))
+		controlErr = syscall.Dup2(int(newFD), int(fd))
+		if controlErr != nil {
+			return
+		}
+		// Go normally opens file descriptors with O_CLOEXEC. This prevents races where
```
fredericdalleau: I doubt `syscall.Dup2` would set the `O_CLOEXEC` on the new fd. Do you have a reference for this comment?
Have you considered `syscall.Dup3` ? Though I have no idea how it behaves on mac.

djs55: Ah sorry I should have squashed this commit -- the code is deleted in a later patch. I had a look on Mac for `Dup3` but unfortunately they don't have it :(

fredericdalleau (approved):

PR #538, opened 2021-05-07
transport: allow named pipe SecurityDescriptor to be set

djs55:
This is needed on some Windows machines to set the ACLs on the named pipes to ensure we can connect to them.

### DIFF
--- a/go/pkg/vpnkit/transport/transport.go
+++ b/go/pkg/vpnkit/transport/transport.go
@@ -10,6 +10,10 @@ type Transport interface {
 	Dial(_ context.Context, path string) (net.Conn, error)
 	Listen(path string) (net.Listener, error)
 	String() string
+	// SetSecurityDescriptor for Windows named pipes in SDDL format, see
+	// https://docs.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-definition-language
+	// For example to allow Administrator and System: "D:P(A;;GA;;;BA)(A;;GA;;;SY)".
+	SetSecurityDescriptor(sddl string)
 }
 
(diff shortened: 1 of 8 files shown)


go/pkg/vpnkit/transport/transport.go
```diff
@@ -10,6 +10,8 @@ type Transport interface {
 	Dial(_ context.Context, path string) (net.Conn, error)
 	Listen(path string) (net.Listener, error)
 	String() string
+	// SetSecurityDescriptor for Windows named pipes.
+	SetSecurityDescriptor(sddl string)
```
fredericdalleau, line 14: I don't see where you call this function?
Why not simply use a default security descriptor?

djs55: It's a bit of an awkward split but this library is vendored inside the Docker Desktop codebase and a Desktop-specific descriptor is set which references the group `docker-users` which is created by the  installer. I was hoping to avoid adding a dependency on the Desktop installer here.

Perhaps we could change the API so the client provides a "listener factory" (not sure what would be idiomatic in Go) which would encapsulate this on the client-side.

fredericdalleau: I see, maybe you can add an exemple security descriptor as a comment, and also make sure it stills works as before if the sddl is not set

djs55: Thanks -- I've expanded the comment with a link and an example. I've added a simple unit test to demo that it still works without the descriptor being set.

fredericdalleau (approved):
LGTM thanks

PR #539, opened 2021-05-10
Replace `Uwt` with `Luv`

djs55:
Both [Uwt](https://github.com/fdopen/uwt) and [Luv](https://aantron.github.io/luv/luv/index.html) are bindings to [libuv](https://libuv.org), the scalable IO library used by [node.js](https://nodejs.org/en/).

Advantages of `Uwt`:
- it is very stable and we've tested it thoroughly in production
- it exposes an `Lwt`-based API for promises so works easily with the rest of our code

Disadvantages of `Uwt`:
- It hasn't been worked on in 3 years see https://github.com/fdopen/uwt/pull/5 
- it uses an old build tool ([omake](https://github.com/ocaml-omake/omake)) which prevents us from switching everything to [dune](https://github.com/ocaml/dune), removing Cygwin to fix our Windows build and (optionally) vendoring our dependencies with [opam monorepo](https://github.com/ocamllabs/opam-monorepo).

Advantages of `Luv`:
- it's likely to become the IO foundation for `Lwt` in future (`Luv` and `Lwt` have the same maintainers)
- it builds with `dune` so will help us remove Cygwin, fix our Windows build and vendor our dependencies

Disadvantages of `Luv`:
- it exposes the raw callback API of `libuv` so we need to carefully interface with the `Lwt` event loop
- we haven't tested it as thoroughly

`vpnkit` had already separated out IO into a separate module, which was needed to migrate from `Lwt_unix` to `Uwt` in the past. However the `Luv` interface is a thinner layer on top of `libuv` which means we need to carefully interface the `libuv` callback-based world with the `Lwt` promise world. The proposal here is to run `Luv` on one thread and `Lwt` on another, and to use Mutex-protected Queues to send work between the two event loops.

A typical function which can be called from `Lwt` looks like this:

```ocaml
let do_some_io () : unit Lwt.t =
    (* Called from the Lwt universe, which is where the TCP/IP stack runs *)
    Luv_lwt.in_luv (fun return ->
        (* Now we're in the Luv event loop *)
        Luv.Do.something _ (function
            (* libuv calls callbacks when interesting things happen.
               We want to tell the Lwt caller on the other thread.
               The helper function `return` pushes a result on the return queue. *)
            | Error err -> return (Error (`Msg (Luv.Error.strerror err)))
            | Ok () -> return (Ok ())
        )
    )
```
where `Luv_lwt.in_luv` pushes the function to a queue and signals the `Luv` event loop, and then when a return result is finally ready, the function `return` pushes the value to a queue and signals the `Lwt` event loop. The queues and signalling is all in a small [module Luv_lwt](https://github.com/moby/vpnkit/pull/539/files#diff-59fbd756155375341a3d58d5168f420638cedc52feeeab447a06a2c6af746c8dR110).

The callback code is bit nested in places in this style:

```ocaml
match foo () with
| Error _ ->
| Ok () ->
  match bar () with
  | Error _ ->
```

see for example [binding a UDP socket](https://github.com/moby/vpnkit/pull/539/files#diff-cf10f0a935e147371e1f96c97714977aa7ef78daa153f4f44d65f98751972b77R240) (you have to expand the collapsed diff of `host.ml` otherwise it's not visible).

but I thought I'd write everything out before trying to be too clever with combinators.

This was a good opportunity to start writing some inline unit tests to complement the existing end-to-end tests.

It was also a good opportunity to reformat the new code with `ocamlformat`.


src/bin/connect.ml
```diff
@@ -71,11 +71,14 @@ module Hvsock = struct
...
   | Some sockaddr ->
     let description = "hvsock" in
-    Host.Sockets.register_connection description >>= fun idx ->
-    let fd = F.Socket.create () in
-    F.Socket.connect fd sockaddr >|= fun () ->
-    let flow = F.connect fd in
-    { idx; flow }
+    begin match Connection_limit.register description with
+    | Error (`Msg m) -> Lwt.fail_with m
+    | Ok idx ->
```
avsm, line 76: This looks a bit like `Lwt_result.get_exn` but specialised for the Msg case -- https://ocsigen.org/lwt/latest/api/Lwt_result

Perhaps define a quick operator for that and use it in all the places you do the Error -> Lwt.fail_with conversion?

djs55: I added a
```ocaml
let ( >>*= ) m f = m >>= function
  | Error (`Msg m) -> Lwt.fail_with m
  | Ok x -> f x
```
and used it in a bunch of places.

src/hostnet/hosts.mli
```diff
@@ -11,12 +11,12 @@ module Make(Files: Sig.FILES): sig
 
   type watch
 
-  val watch: ?path:string -> unit -> (watch, [ `Msg of string ]) result
+  val watch: ?path:string -> unit -> (watch, [ `Msg of string ]) result Lwt.t
```
avsm, line 14: Not quite clear what the Lwt.t here does. Does it block until the watching begins, or does it block indefinitely until the watch terminates?

avsm (approved):
LGTM; just a minor question about clarifying `watch` (and the associated Lwt.async when calling it). Will be more important when porting to eio.

src/hostnet/hosts.mli
djs55: It blocks until the watching begins. I should clarify that in a doc, as it's not very obvious.

PR #543, opened 2021-05-16
Update all the OCaml dependencies

djs55:
On top of #542 (vendoring the DNS code)

This PR updates the Mirage dependencies to almost the latest versions. It requires a few more  pins for now:
- unreleased `hvsock.3.0.0`
- unreleased `protocol-9p` and `protocol-9p-unix`

Mostly the changes were mechanical e.g.
- `Mirage_foo_lwt` -> `Mirage_foo`
- `Ipaddr` and `Macaddr` use result types
- there is no `Clock.t` any more
- new libraries for `Arp`, `Ethernet`

Some of the changes were trickier:
- The `write` functions now take a `fill` function which is supplied a buffer
- The handling of the source IP address in the IP stack is different: it uses a `cidr` and a `gateway`


mat007 (approved):
It’s green!

PR #544, opened 2021-05-23
WIP: re-enable appveyor for Windows vpnkit.exe builds

djs55:
This is based on #539

It builds locally for me with `luv` rather than `uwt`. I also found some advice to use `apt-cyg` to install cygwin packages. There seems to be an opam 2 (`opam-installer`) and an `ocaml` 4.10 package already.

If this works then I think I'll rename `Makefile.darwin` into `Makefile` for all 3 platforms.


djs55:
Appveyor  is working again so this PR can be closed.

PR #546, opened 2021-07-12
Ignore a dup SYN for older connections as well

yamt:
I occasionally observed stalls because of this
when using NuttX sim as a client.


djs55:
@yamt thanks for the patch! I think I might have seen a similar problem.

I'll test it and let you know what I find. Unfortunately the CI isn't fully enabled on this repo at the moment.

yamt:
i applied this to our private ci two months ago.
i haven't noticed any issues since then.
before applying this patch, it was occasionally failing. (once a few days or so)

djs55:
@yamt thanks for the update. I'll merge the patch now.

PR #552, opened 2022-01-17
Update dependencies again, attempt to re-enable appveyor

djs55:
Update all the dependencies to latest, mostly minor changes:
- `Cstruct.len` is renamed `Cstruct.length`
- `Fmt.strf` is renamed `Fmt.str`
- `mirage-protocols` has been deprecated in favour of using libraries directly (`Tcpip`, `Arp`), removing a layer of indirection

The most significant change is the removal of the `~listeners` argument on the TCP `input`. This has been replaced with a `listen` and `unlisten`. Currently I've not added any calls to `unlisten` so it's probably leaking.

I've added an `appveyor.yml` based on mirage/mirage-block, to see if the CI build can be restored.

The minimum OCaml version needed by dependencies is now 4.08.

The remaining Windows package pins correspond to open PRs which I'd like to merge and release as soon as possible.


djs55:
CI is green! Let's merge this to make `master` green too!

PR #553, opened 2022-01-21
appveyor: enable tests, build artifacts

djs55:
- re-enable tests
- use `dune` rules more to reduce dependencies on make, shell, cygwin on Windows
- update the old license file generation to use .json


djs55:
CI is green and the artefacts are all present!

PR #555, opened 2022-01-24
licenses: include name and license type (e.g. SPDX)

djs55:


djs55:
Looking better! Merging

PR #556, opened 2022-01-25
Don't calculate high_ip in DHCP configuration and zap outdated comment

haesbaert:
This is a left over from when charrua would only work if there was a delta of at
least one between min and max addresses.

Also zap two outdated comments related to the same issue.

I can't really test this code since I have no environment so please be careful.

Relevant commits
https://github.com/mirage/charrua/commit/3c1995bab9679c0105c962d269bec92806868ffa
https://github.com/mirage/charrua/commit/0bb4464f41dab7fadc37bdda606fe76edaefc6d3


djs55:
The CircleCI build for unrelated reasons (a pinned OCaml branch has been deleted -- I need to update it to a sha in the master branch)

djs55 (approved):
This makes sense and it's always good to actually resolve a FIXME!

LGTM

djs55:
I'll fix CircleCI separately and then merge this.

djs55:
CircleCI is fixed on the master branch, I'll merge this.

PR #557, opened 2022-01-29
opam: add pin-depends from the Dockerfile to get the solver to pass

avsm:


avsm:
Needs https://github.com/moby/vpnkit/pull/539 first, which @djs55 is doing. Closing this until then.

PR #559, opened 2022-02-16
Simplify build instructions

djs55:
We don't need to pin unreleased versions of hvsock, protocol-9p since these have been released upstream.


djs55:
CircleCI and Appveyor are green.

PR #561, opened 2022-03-09
Update go mod to 1.17

mat007:
Changed `1.16` to `1.17` then did a `go mod tidy`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,44 +1,68 @@
 module github.com/moby/vpnkit
 
-go 1.16
+go 1.17
 
 require (
 	github.com/Microsoft/go-winio v0.4.11
-	github.com/PuerkitoBio/purell v1.1.0 // indirect
-	github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 // indirect
-	github.com/emicklei/go-restful v2.4.0+incompatible // indirect
-	github.com/emicklei/go-restful-swagger12 v0.0.0-20170208215640-dcef7f557305 // indirect
-	github.com/go-openapi/jsonpointer v0.0.0-20170102174223-779f45308c19 // indirect
-	github.com/go-openapi/jsonreference v0.0.0-20161105162150-36d33bfe519e // indirect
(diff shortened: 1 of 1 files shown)


djs55 (approved):
LGTM!

go.mod
```diff
@@ -1,44 +1,68 @@
...
-go 1.16
+go 1.17
 
 require (
 	github.com/Microsoft/go-winio v0.4.11
+	github.com/google/uuid v0.0.0-20170814143639-7e072fc3a7be
+	github.com/labstack/echo v3.3.10+incompatible
+	github.com/linuxkit/virtsock v0.0.0-20180830132707-8e79449dea07
+	github.com/pkg/errors v0.8.1-0.20181023235946-059132a15dd0
+	github.com/sirupsen/logrus v1.3.0
```
thaJeztah, line 11: we should update this to a more current version; IIRC, this old version is known to not be compatible with go1.13 and up

go.mod
```diff
@@ -1,44 +1,68 @@
...
 
-go 1.16
+go 1.17
 
 require (
 	github.com/Microsoft/go-winio v0.4.11
+	github.com/google/uuid v0.0.0-20170814143639-7e072fc3a7be
+	github.com/labstack/echo v3.3.10+incompatible
+	github.com/linuxkit/virtsock v0.0.0-20180830132707-8e79449dea07
+	github.com/pkg/errors v0.8.1-0.20181023235946-059132a15dd0
```
thaJeztah, line 10: This version is not compatible with go1.13 error wrapping; we should update that as well

go.mod
```diff
@@ -1,44 +1,68 @@
 module github.com/moby/vpnkit
 
-go 1.16
+go 1.17
 
 require (
 	github.com/Microsoft/go-winio v0.4.11
+	github.com/google/uuid v0.0.0-20170814143639-7e072fc3a7be
```
thaJeztah, line 7: this one has tagged releases, and (I may be mistaken ) but there may be security issues with older version

PR #563, opened 2022-03-14
vpnkit-forwarder: avoid -data-connect spinning

djs55:
Previously if -data-connect was connecting to a port forward and the remote isn't running, it would connect but then get an immediate EOF. Unfortunately it would try again and spin in a tight loop. This is likely to happen when shutting down, as the remote can stop before the port forward closes.

This patch adds a one second delay to break the spinning and a log every 30s to highlight any persistent problem.

### DIFF
--- a/scripts/licenses.ml
+++ b/scripts/licenses.ml
@@ -601,7 +601,7 @@ Copyright (c) 2009-2018 <the authors, see individual headers on files>
 
     |} ^ isc
   }
-  | "functoria-runtime.3.0.3" -> {
+  | "functoria-runtime.3.0.3" | "functoria-runtime.3.1.2" -> {
     link = "https://github.com/mirage/functoria/blob/d89fadb14bef0b1eb43761cdced0816e7772a533/LICENSE.md";
     text = {|
     Copyright (c) 2015-2018, Thomas Gazagnaire, Anil Madhavapeddy, Dave Scott, Thomas Leonard, Gabriel Radanne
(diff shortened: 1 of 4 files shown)


scripts/licenses.ml
```diff
@@ -601,7 +601,7 @@ Copyright (c) 2009-2018 <the authors, see individual headers on files>
 
     |} ^ isc
   }
-  | "functoria-runtime.3.0.3" -> {
+  | "functoria-runtime.3.0.3" | "functoria-runtime.3.1.2" -> {
```
mat007, line 604: (curious)
How do you notice the licenses need to be updated? How do you do the update?

mat007 (approved):
Makes sense!

scripts/licenses.ml
djs55: It's not a great system. The appveyor build fails and gives me a list, mostly consisting of new minor versions. I manually take a look to verify that they haven't changed. 99.99% of the time nothing changes so it's a bit tedious. It would be nice to automate it a bit more. In my ideal world I would just review any changed license text.

Maybe if I had a tool which could go from github tag link -> license data I could use it to update a checked-in cache of licenses.

mat007: > Maybe if I had a tool which could go from github tag link -> license data I could use it to update a checked-in cache of licenses.

Oh, I do have that elsewhere!
I can extract a command from it, what would you need? Would that work for instance:
```
./pinata.sh license info https://github.com/bla/blo
```
?

PR #568, opened 2022-04-05
Add missing licenses

mat007:
According to https://ci.appveyor.com/project/moby/vpnkit/builds/42899712 they are a couple of licenses missing from `licenses.ml`:
```
Fatal error: exception Failure("unknown licenses for mmap.1.2.0, camlp-streams.5.0")
```

Given that the CircleCI build passed for the same commit, I suppose Windows and Mac don’t pull the same set of dependencies?


mat007:
Oh OK #567 was taking care of it already!

PR #569, opened 2022-04-05
Update licenses versions

mat007:
https://github.com/mirage/mirage-channel/releases/tag/v4.1.0 was released a few hours ago.

I’m still a bit puzzled as to why the dependencies aren’t pinned. What did I miss? 🤔

### DIFF
--- a/scripts/licenses.ml
+++ b/scripts/licenses.ml
@@ -580,8 +580,9 @@ POSSIBILITY OF SUCH DAMAGE.
 Copyright (c) 2009-2018 <the authors, see individual headers on files>
     |} ^isc
   }
-  | "cstruct.6.0.1" | "cstruct-lwt.6.0.1" | "cstruct-sexp.6.0.1" | "ppx_cstruct.6.0.1" -> {
-    link = "https://github.com/mirage/ocaml-cstruct/blob/28dade8963a9edfddeb8dba782a8eae0341e80d4/LICENSE.md";
+  | "cstruct.6.0.1" | "cstruct-lwt.6.0.1" | "cstruct-sexp.6.0.1" | "ppx_cstruct.6.0.1"
+  | "cstruct.6.1.0" | "cstruct-lwt.6.1.0" | "cstruct-sexp.6.1.0" | "ppx_cstruct.6.1.0" -> {
+    link = "https://raw.githubusercontent.com/mirage/ocaml-cstruct/v6.1.0/LICENSE.md";
     text = {|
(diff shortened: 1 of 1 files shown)


djs55 (approved):
LGTM
Yeah we probably should pin some of these again.

mat007:
CircleCI build doesn’t agree that the versions need to be bumped
> Fatal error: exception Failure("unknown licenses for ppx_cstruct.6.0.1, mirage-channel.4.0.1, cstruct-sexp.6.0.1, cstruct-lwt.6.0.1, cstruct.6.0.1")

🤔

djs55:
Hm, not sure if that's because of non-determinism in the package manager's SAT solver (😬 )or something else. I'll take a look at pinning.

scripts/licenses.ml
```diff
@@ -580,8 +580,8 @@ POSSIBILITY OF SUCH DAMAGE.
 Copyright (c) 2009-2018 <the authors, see individual headers on files>
     |} ^isc
   }
-  | "cstruct.6.0.1" | "cstruct-lwt.6.0.1" | "cstruct-sexp.6.0.1" | "ppx_cstruct.6.0.1" -> {
-    link = "https://github.com/mirage/ocaml-cstruct/blob/28dade8963a9edfddeb8dba782a8eae0341e80d4/LICENSE.md";
+  | "cstruct.6.1.0" | "cstruct-lwt.6.1.0" | "cstruct-sexp.6.1.0" | "ppx_cstruct.6.1.0" -> {
```
djs55, line 583: To keep it working on CircleCI, you could keep both of these sets of versions e.g.
```ocaml
| "cstruct.6.0.1" | "cstruct-lwt.6.0.1" | "cstruct-sexp.6.0.1" | "ppx_cstruct.6.0.1" | "cstruct.6.1.0" | "cstruct-lwt.6.1.0" | "cstruct-sexp.6.1.0" | "ppx_cstruct.6.1.0" -> {
```

mat007:
> or something else

Maybe there is a diamond style dependency with both versions still being needed.

scripts/licenses.ml
mat007: Yes, good point, I’ll do that!

djs55:
As an experiment I've added version pins for the direct dependencies in this PR: https://github.com/moby/vpnkit/pull/570

There may be indirect dependencies too but I don't have a nice way to pin those, except by making them direct dependencies. I suspect it's the direct dependencies that change the most often (like the ones in this PR).

djs55:
CI is green on Mac and Win (ocaml-ci is known to be broken)

PR #571, opened 2022-04-05
Fix typo in licenses

mat007:
This silently adds `fd-send-recv.2.0.1` with an empty type.

### DIFF
--- a/scripts/licenses.ml
+++ b/scripts/licenses.ml
@@ -905,7 +905,7 @@ might be covered by the GNU Lesser General Public License.
   }
   | "fd-send-recv.2.0.1" -> {
     link = "https://github.com/xapi-project/ocaml-fd-send-recv/blob/7c9b151a7cc54f11c45280177027d9d42473737c/LICENSE";
-    text = lgpl21_with_ocaml_linking_exception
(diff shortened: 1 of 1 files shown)


mat007:
Or maybe it’s due to something else, it didn’t seem to have made a difference…

mat007:
I’m not sure why else, these now look similar
```
  | "fd-send-recv.2.0.1" -> {
    link = "https://github.com/xapi-project/ocaml-fd-send-recv/blob/7c9b151a7cc54f11c45280177027d9d42473737c/LICENSE";
    text = lgpl21_with_ocaml_linking_exception;
  }
```
and 
```
  | "ocplib-endian.1.2" -> {
    link = "https://github.com/OCamlPro/ocplib-endian/blob/10292cd3ffa4d23d737e3f855ad04f22d3d95460/COPYING.txt";
    text = lgpl21_with_ocaml_linking_exception;
  }
```
but the latter has a correct `type` inferred…

djs55 (approved):
I'm a bit rusty, but I thought the trailing `;` was optional. However it makes for cleaner `diff` so I'm all for it!

mat007:
Oh I think it is optional. I can’t understand why computing the type works for some and not exactly identical others…

PR #572, opened 2022-04-06
appveyor: ensure we generate the COMMIT artefact

djs55:
Previous builds seemed to have a cached COMMIT file. Dune doesn't have a "PHONY" concept as far as I can see, perhaps it was a mistake to move this code in there.


djs55:
I think it's actually fine, the "problem" is that appveyor builds a merge commit, not the PR commit.

PR #580, opened 2022-04-25
Unlisten sockets after they disconnect

fredericdalleau:
After a successful connection forwarding to a server, if the server
disappears, a tentative to connect the server again will be accepted by
vpnkit even if vpnkit failed to reach the server.
Ensure this does not occur by calling unlisten.
Add test for valid and invalid connection cases

Signed-off-by: Frédéric Dalleau <frederic.dalleau@docker.com>

### DIFF
--- a/src/hostnet_test/test_half_close.ml
+++ b/src/hostnet_test/test_half_close.ml
@@ -166,6 +166,77 @@ let test_host_half_close () =
       )
   end
 
+let test_connect_valid_invalid_port () =
+  Host.Main.run begin
+    Slirp_stack.with_stack ~pcap:"test_connect_valid_invalid_port.pcap" (fun _ stack -> with_server (fun _ ->
+        Lwt.return_unit
+      ) (fun server ->
+            (* Now that a server is running, connect to a valid port and ensure it succeeds quickly *)
+            let open Slirp_stack in
+            let ip = Ipaddr.V4.localhost in
+            let port = server.Server.port in
+            let mkconn = Client.TCPV4.create_connection (Client.tcpv4 stack.t) (ip, port)
+            >|= function
+            | Ok _ ->
+              Log.debug (fun f ->
+                  f "Connected to localhost:%d" port);
+            | Error _ ->
+              Log.err (fun f ->
+                  f "Failure to connect to localhost:%d" port);
+              failwith "Connection should have succeeded";
+            >>= fun () ->
+              Server.destroy server;
+            >>= fun () ->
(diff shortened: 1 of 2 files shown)


src/hostnet_test/test_half_close.ml
```diff
@@ -166,6 +166,105 @@ let test_host_half_close () =
...
 
+let test_connect_invalid_port () =
+  Host.Main.run begin
+    Slirp_stack.with_stack ~pcap:"test_connect_invalid_port.pcap" (fun _ stack -> with_server (fun _ ->
+        Lwt.return_unit
+      ) (fun server ->
+            (* Now that a server is running, connect to an invalid port and ensure it fails quickly *)
+            let open Slirp_stack in
+            let ip = Ipaddr.V4.localhost in
+            let port = server.Server.port + 1 in
```
djs55, line 177: Perhaps on the Windows machine `server.Server.port + 1` was used?
```
**** Starting test TCP: test server invalid port
...
9.09694 [ERROR] Connected to localhost:1986
```
Maybe quickly allocating and deallocating a port will be more reliable, like
```ocaml
with_server (fun _ -> Lwt.return_unit) (fun server -> Server.destroy server >>= fun () -> Lwt.return server.Server.port)
>>= fun invalid_port ->
```

fredericdalleau: Probably, in that case, it becomes pretty similar than the next test which ensure connection succeeds first and then destroys the server

djs55:
Regarding the Windows failure: it might be a Windows firewall issue on the test running machine.  Perhaps the test should be gated with something like
```ocaml
if Sys.os_type = "Win32" then begin
  ...
end else begin
  ...
end
```
Or maybe a timeout problem? This Go program:
```go
package main

import (
  "fmt"
  "net"
  "time"
)

func main(){
  start := time.Now()
  _, err := net.Dial("tcp", "127.0.0.1:8080")
  if err != nil {
    fmt.Printf("rejected after %s\n", time.Since(start))
    return
  }
  fmt.Printf("connected after %s\n", time.Since(start))
}
```
takes >2s:
```
PS C:\Users\dave\go\src\connect-localhost> go build
PS C:\Users\dave\go\src\connect-localhost> .\connect-localhost.exe
rejected after 2.0195789s

djs55 (approved):
LGTM (and Mac and Windows CI are green!)

yamt:
i feel lucky today because i hit this bug today and it has already been fixed. thank you.h

PR #582, opened 2022-05-17
libproxy: allow read/write buffers to be set

djs55:
If the link has high latency then it's useful to use larger buffers

### DIFF
--- a/go/pkg/libproxy/multiplexed.go
+++ b/go/pkg/libproxy/multiplexed.go
@@ -337,12 +337,19 @@ type Multiplexer interface {
 	Run()            // Run the multiplexer (otherwise Dial, Accept will not work)
 	IsRunning() bool // IsRunning is true if the multiplexer is running normally, false if it has failed
 
-	Dial(d Destination) (Conn, error)    // Dial a remote Destination
-	Accept() (Conn, *Destination, error) // Accept a connection from a remote Destination
+	Dial(d Destination) (MultiplexedConn, error)    // Dial a remote Destination
+	Accept() (MultiplexedConn, *Destination, error) // Accept a connection from a remote Destination
 
 	Close() error // Close the multiplexer
 
 	DumpState(w io.Writer) // WriteState dumps debug state to the writer
+}
+
+// MultiplexedConn is a sub-connection within a single multiplexed connection.
+type MultiplexedConn interface {
+	Conn
+	SetReadBuffer(uint) error  // SetReadBuffer sets the maximum read buffer size
+	SetWriteBuffer(uint) error // SetWriteBuffer sets the maximum write buffer size
 }
 
 type multiplexer struct {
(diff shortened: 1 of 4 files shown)


go/pkg/libproxy/multiplexed.go
```diff
@@ -14,29 +14,28 @@ import (
...
 }
 
 func (w *windowState) isAlmostClosed() bool {
-	return w.size() < maxBufferSize/2
+	return w.size() < int(w.max/2)
 }
 
 func (w *windowState) advance() {
-	w.allowed = w.current + uint64(maxBufferSize)
+	w.allowed = w.current + w.max
```
fredericdalleau, line 38: I wonder how long it takes to overlap?

go/pkg/libproxy/multiplexed.go
```diff
@@ -14,29 +14,28 @@ import (
...
+	return fmt.Sprintf("current %d, allowed %d, max %d", w.current, w.allowed, w.max)
 }
 
 func (w *windowState) size() int {
 	return int(w.allowed - w.current)
 }
 
 func (w *windowState) isAlmostClosed() bool {
-	return w.size() < maxBufferSize/2
+	return w.size() < int(w.max/2)
```
fredericdalleau, line 34: What if w.max is set to an odd value ?

djs55: Normally it doesn't send window updates (unlike TCP), it only sends one when `isAlmostClosed`. If `max` was odd, then `max / 2` would round down, and it would send a window update 1 byte earlier. To check nothing bad happens, I've pushed a patch to set the max to `defaultWindowSize-1` in one of the test cases.

go/pkg/libproxy/multiplexed.go
djs55: I think if it went at (a very optimistic) 8Gbit/sec = 1 GiB/sec = `2 ** 30` bytes per second, I think it will overflow to `0` in `2 ** (64-30)` seconds which is > 500 years
```python
>>> 2**(64-30) / 60 / 60 / 24 / 365
544.7700781329274
``` 
I hope we'll have replaced the code by then! (As an aside, I'm interested in any other multiplexing libraries we could use. There was a SPDY implementation but I think it's deprecated. Perhaps HTTP/2 can do what we want? Or maybe even send packets over gRPC? I'm not sure if the overheads would be too much though.)

go/pkg/libproxy/multiplexed.go
```diff
@@ -249,6 +252,26 @@ func (c *channel) recvClose() {
 	c.c.Broadcast()
 }
 
+// SetReadBuffer sets the size of the operating system's receive buffer associated with the connection.
+// See similar function https://pkg.go.dev/net#IPConn.SetReadBuffer
+func (c *channel) SetReadBuffer(bytes int) error {
```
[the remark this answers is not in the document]
fredericdalleau: Ha I understand why it was `int` it can be setsockopt, it's bounded and can fail.

SO_RCVBUF
              Sets or gets the maximum socket receive buffer in bytes.
              The kernel doubles this value (to allow space for
              bookkeeping overhead) when it is set using [setsockopt(2)](https://man7.org/linux/man-pages/man2/setsockopt.2.html),
              and this doubled value is returned by [getsockopt(2)](https://man7.org/linux/man-pages/man2/getsockopt.2.html).  The
              default value is set by the
              /proc/sys/net/core/rmem_default file, and the maximum
              allowed value is set by the /proc/sys/net/core/rmem_max
              file.  The minimum (doubled) value for this option is 256.
boucane:~$ cat /proc/sys/net/core/rmem_max
212992

fredericdalleau (approved):
LGTM

PR #583, opened 2022-06-01
libproxy: ability to Close(), expose NewLoopback()

djs55:
- `Close()` allows us to cleanly shutdown the multiplexer (similar to `Close()` on a `net.Listener`)
- `NewLoopback()` satisfies the interface of the multiplexer, which makes it easier to write test cases in downstream projects


fredericdalleau (approved):
LGTM. I don't see a test that closes the mux though.

PR #584, opened 2022-06-04
metadata fixes for opam/ocaml versions

avsm:
This adds pin-depends to the opam file to make it easier to build with simpler opam invocations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 _build/
+_opam/
 .*.swp
 *.native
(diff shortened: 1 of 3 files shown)


avsm:
I think it's better to switch to `luv` from `uwt` sooner rather than later... the compilation of omake is challenging now.

djs55 (approved):
LGTM!

djs55:
Also builds fine for me locally.

PR #590, opened 2022-07-27
Add a mechanism to forward outgoing TCP via a Unix domain socket

djs55:
This is based on the existing "gateway forwards" mechanism which allowed traffic sent to the gateway to be forwarded.

A `forwards.json` can be dynamically updated with IP network matches and Unix domain socket / Windows named pipe paths. When a SYN arrives, a request is sent on the Unix domain socket, allowing the server to reject or accept the request. Assuming the connection is accepted, the handshake is completed and traffic flows.

### DIFF
--- a/src/hostnet/forwards.ml
+++ b/src/hostnet/forwards.ml
@@ -416,3 +416,72 @@ module Stream = struct
       | `Forwarded flow -> Forwarded.shutdown_read flow
   end
 end
+
+module Test (Clock : Mirage_clock.MCLOCK) = struct
+  module Remote = Read_some (Host.Sockets.Stream.Unix)
+
+  module Proxy =
+    Mirage_flow_combinators.Proxy (Clock) (Remote) (Host.Sockets.Stream.Tcp)
+
+  module Handshake = Handshake (Remote)
+  open Lwt.Infix
+
+  type server = Host.Sockets.Stream.Unix.server
+
+  let start_forwarder path =
+    Host.Sockets.Stream.Unix.bind path >>= fun s ->
+    Host.Sockets.Stream.Unix.listen s (fun flow ->
+        Log.info (fun f -> f "accepted flow");
+        let local = Remote.connect flow in
+        Handshake.Request.read local >>= function
+        | Error e ->
+            Log.info (fun f ->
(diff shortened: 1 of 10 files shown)


src/hostnet/forwards.ml
```diff
@@ -0,0 +1,487 @@
...
+  (* A Message is a buffer prefixed with a uint16 length field. *)
+  module Message = struct
+    open Lwt.Infix
+
+    let pp_error ppf = function
+      | `Flow e -> FLOW.pp_error ppf e
+      | `Eof -> Fmt.string ppf "EOF while reading handshake"
+
+    let read flow =
+      FLOW.read_some flow 2 >>= function
```
avsm, line 152: minor, but I think you could replace this `>>=` with `>!=` and lose all the `Lwt.return`s below.

src/hostnet/forwards.ml
```diff
@@ -0,0 +1,487 @@
...
+    }
+
+    open Ezjsonm
+
+    let of_json j =
+      let protocol = Protocol.of_json @@ find j [ "protocol" ] in
+      let src_ip =
+        match Ipaddr.of_string @@ get_string @@ find j [ "src_ip" ] with
+        | Error (`Msg m) ->
+            raise (Parse_error (j, "src_ip should be an IPv4 address: " ^ m))
```
avsm, line 185: Doesn't `Ipaddr.of_string` also parse V6 addresses? Dont you need `Ipaddr.V4.of_string` here?

src/fs9p/dune
```diff
@@ -1,4 +1,4 @@
 (library
  (name fs9p)
  (wrapped false)
- (libraries protocol-9p mirage-flow))
+ (libraries protocol-9p mirage-flow result))
```
avsm, line 4: I don't think you need this `result` package if you use a modern OCaml. 4.08+ (as the lower bound currently is) should be sufficient as `Result` has been present since then.

avsm (approved):
lgtm just minor comments

src/hostnet/forwards.ml
djs55: Ah, I think in this case the accompanying error message is wrong -- I've been trying to use `Ipaddr.t` more throughout in preparation for supporting IPv6

PR #593, opened 2022-08-14
Ensure uname subprocess is cleaned up

### LINKED ISSUE #592: Resource leak spawning uname subprocess call on Darwin
The `uname` command / subprocess spawned by vpnkit is not cleaned up / reaped appropriately. The result is a zombie subprocess which lingers in the process table.

This seems to be due to not calling `Unix.close_process_in` instead of `close_in` in `src/bin/config/discover.ml`

Screenshot of `htop`

Screen Shot 2022-08-14 at 5 24 14 pm

Output of `ps -o pid,state,xstat,command -p 72041`
```
  PID STAT XSTAT COMMAND
72041 Z        0 (uname)
```

tmoschou:
Adds a fix to the `uname` command / subprocess spawned which was not cleaned up / reaped appropriately on Darwin.
The result was a zombie process / resource leak that lingered in the process table.

Fixes #592

### DIFF
--- a/src/bin/config/discover.ml
+++ b/src/bin/config/discover.ml
@@ -4,7 +4,7 @@ let os_type () =
     (* We want to differentiate between Linux and Darwin/macOS *)
     let ic = Unix.open_process_in "uname" in
     let line = input_line ic in
-    close_in ic;
+    let _ = Unix.close_process_in ic in
     line
   | x -> x
 


tmoschou:
Looks like CircleCI is using Xcode 12.0.0 which is no longer supported 
https://circleci.com/docs/xcode-policy#xcode-image-retention-and-deprecation
https://circleci.com/docs/using-macos#supported-xcode-versions

djs55 (approved):
Thanks, the change looks fine to me. I'll investigate the CI.

I'm slightly surprised that this code ended up in the final binary. I think it is only supposed to be used at build-time. Maybe it needs to move out of `src/bin`?

djs55:
I've taken the liberty of force-pushing a rebase after the recent CI improvements.

djs55:
OK, CI is a bit greener :)

PR #594, opened 2022-08-22
circleci: bump to latest xcode, macOS 10.15

djs55:
macOS 10.15 is Catalina+ and matches the current Docker Desktop limits https://docs.docker.com/desktop/install/mac-install/


djs55:
CircleCI is better, so merging

PR #603, opened 2022-10-07
dhcp: offer very long leases

djs55:
Previously users would see network glitches every 2hrs as the lease was renewed.

Since vpnkit is used in "static" environments, we can offer very long leases. This value is in seconds and > 68 years.


avsm (approved):
DHCP does allow  `0xffffffff` to represent an infinite lease too, which might be a bit cleaner.

PR #612, opened 2022-11-09
tweak some of the instructions for setting up on a new machine

nicks:
- The makefile has two 'depends' targets, and it was ignoring the first one
- also requires `dylibbundler`
- got some license errors so i fixed them


djs55 (approved):
Thanks!

PR #619, opened 2023-02-17
go: avoid possible overflow of the Write window

djs55:
The Write window is used to keep track of how much buffer space is free in the remote to avoid one connection blocking the rest. Previously we checked the window and decided how much to write, then dropped the metadata mutex before performing the write. In theory another Write call on the same connection could see that buffer size is free, send too much and block the connection.

Therefore Write should take ownership of the space by bumping the `current` window before dropping the lock.

### DIFF
--- a/go/pkg/libproxy/multiplexed.go
+++ b/go/pkg/libproxy/multiplexed.go
@@ -138,10 +138,14 @@ func (c *channel) Write(p []byte) (int, error) {
 			return written, io.EOF
 		}
 		if c.write.size() > 0 {
+			// Some window space is available
 			toWrite := c.write.size()
 			if toWrite > len(p) {
 				toWrite = len(p)
 			}
+			// Advance the window before dropping the lock in case another Write() appears and sees the same available space
+			c.write.current = c.write.current + uint64(toWrite)
+
 			// Don't block holding the metadata mutex.
 			// Note this would allow concurrent calls to Write on the same channel
 			// to conflict, but we regard that as user error.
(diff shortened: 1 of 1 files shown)


dgageot (approved):

go/pkg/libproxy/multiplexed.go
```diff
@@ -138,10 +138,14 @@ func (c *channel) Write(p []byte) (int, error) {
 			return written, io.EOF
 		}
 		if c.write.size() > 0 {
+			// Some window space is available
 			toWrite := c.write.size()
 			if toWrite > len(p) {
 				toWrite = len(p)
 			}
+			// Advance the window before dropping the lock in case another Write() appears and sees the same available space
+			c.write.current = c.write.current + uint64(toWrite)
```
fredericdalleau, line 147: in case of an error, `c.write.current` will be updated as if the data was sent.
This behavior is different from the previous
You can use a temporary variable instead.

djs55: Unfortunately if there's an error, the whole multiplexer will need to shut down because the messages on the wire will be corrupted. For example if I write:

- Message 1 header: "my payload length is 100 bytes"
- Message 1 payload: only 50 bytes because of an error
- Message 2 header: some other packet

then the remote will think "Message 2 header" is part of "Message 1 payload" and everything breaks :(

So I think in the case of error we need to close everything and report the error upwards.

fredericdalleau: fair enough

fredericdalleau (approved):
LGTM

PR #621, opened 2023-02-17
go: allow concurrent Write calls

djs55:
The [net.Conn interface](https://docker.atlassian.net/browse/POS-1477) says:
> // Multiple goroutines may invoke methods on a Conn simultaneously.

Previously concurrent calls to `Write()` would block for window space but only one would be woken up by a `Signal()` when a window update is received, allowing the other call to block forever.

The handling of write timeouts was also confusing because
- `time.Sleep` couldn't be interrupted
- the condition variable `Wait()` was on a temporary goroutine

### DIFF
--- a/go/pkg/libproxy/multiplexed.go
+++ b/go/pkg/libproxy/multiplexed.go
@@ -99,7 +99,7 @@ func (c *channel) sendWindowUpdate() error {
 	c.read.advance()
 	seq := c.read.allowed
 	c.m.Unlock()
-	return c.multiplexer.send(NewWindow(c.ID, seq))
+	return c.multiplexer.send(NewWindow(c.ID, seq), nil)
 }
 
 func (c *channel) recvWindowUpdate(seq uint64) {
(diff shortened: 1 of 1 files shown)


go/pkg/libproxy/multiplexed.go
```diff
@@ -143,57 +145,41 @@ func (c *channel) Write(p []byte) (int, error) {
...
-			if err2 != nil {
-				return written, err2
-			}
-			if err3 != nil {
-				return written, err3
+
+			if err != nil {
+				return written, err
 			}
 			c.write.current = c.write.current + uint64(toWrite)
```
fredericdalleau, line 155: There is a conflic to resolve here

go/pkg/libproxy/multiplexed.go
```diff
@@ -424,14 +410,22 @@ func (m *multiplexer) appendEvent(e *event) {
...
 	defer m.writeMutex.Unlock()
+	m.appendEvent(&event{eventType: eventSend, frame: f})
+
 	if err := f.Write(m.connW); err != nil {
-		return err
+		return fmt.Errorf("writing frame %s: %w", f, err)
 	}
-	m.appendEvent(&event{eventType: eventSend, frame: f})
-	return m.connW.Flush()
+	if n, err := m.connW.Write(payload); err != nil || n != len(payload) {
```
fredericdalleau, line 422: Shout it return n?

fredericdalleau: Can you write nil?

go/pkg/libproxy/multiplexed.go
```diff
@@ -143,57 +145,41 @@ func (c *channel) Write(p []byte) (int, error) {
...
 			continue
 		}
 
-		// Wait for the write window to be increased (or a timeout)
-		done := make(chan struct{})
-		timeout := make(chan time.Time)
+		// If the client has set a deadline then create a timer:
+		var (
+			timer   *time.Timer
+			timeOut bool
```
fredericdalleau, line 164: Should it be atomic?

fredericdalleau: No, because c.m is already Locked, and is unlock for the duration of c.c.Wait()

go/pkg/libproxy/multiplexed.go
djs55: From friday's discussion: the [Write method docs](https://pkg.go.dev/io#Writer) has
> Write writes len(p) bytes from p to the underlying data stream.

So since `len(nil) == 0` it makes the call a no-op.

also
> Write must return a non-nil error if it returns n < len(p).

So it will return an error to the caller if `n` is too small.

Does this still sound reasonable?

fredericdalleau: yes!

PR #626, opened 2023-03-02
go: don't hold the metadata mutex and then acquire the channel mutex

djs55:
Close() holds the channel mutex then calls decrChannelRef which acquires the metadata mutex.

In DumpState() we should not hold the metadata mutex and then acquire the channel mutex.

### DIFF
--- a/go/pkg/libproxy/multiplexed.go
+++ b/go/pkg/libproxy/multiplexed.go
@@ -574,14 +574,20 @@ func (m *multiplexer) DumpState(w io.Writer) {
 		}
 	})
 	m.eventsM.Unlock()
+
+	var channels []*channel
 	m.metadataMutex.Lock()
-	io.WriteString(w, "Active channels:\n")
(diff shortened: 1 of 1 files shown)


go/pkg/libproxy/multiplexed.go
```diff
@@ -574,14 +574,20 @@ func (m *multiplexer) DumpState(w io.Writer) {
...
 	m.metadataMutex.Lock()
-	io.WriteString(w, "Active channels:\n")
 	for _, c := range m.channels {
+		channels = append(channels, c)
+	}
+	m.metadataMutex.Unlock()
+
+	io.WriteString(w, "Active channels:\n")
+	for _, c := range channels {
 		io.WriteString(w, c.String())
```
mat007, line 587: I can see how having a mutex lock in `channel.String` may be unexpected. 😅

mat007 (approved):

PR #636, opened 2023-05-12
kube: controller lifecycle improvements

milas:
* Optimize the `Dockerfile` to use cache mounts
* Use the `context.Context` during HTTP operations
* Add an internal context for normal controller operations
* Sequence shutdown to wait for informer to stop and then attempt to unexpose any ports before exiting
  * Unexpose ports in parallel, we just log out all the errors regardless


go/cmd/kube-vpnkit-forwarder/main.go
```diff
@@ -1,28 +1,42 @@
...
 	"flag"
-	"log"
 	"os"
 	"os/signal"
 	"syscall"
 	"time"
 
 	"github.com/moby/vpnkit/go/pkg/controller"
 	"github.com/moby/vpnkit/go/pkg/vpnkit"
+	log "github.com/sirupsen/logrus"
```
milas, line 13: `logrus` was being used in the controller but not here - switched for consistency and to add ability to config log level via flag

PR #641, opened 2024-08-06
Add version to package file and pin dependencies to fix build

LaurentGoderre:


LaurentGoderre:
So far, the build succeeds but the e2e tests fail

djs55 (approved):
Since the e2e tests work locally (although not in a container) and we don't care about appveyor (broken -- should be removed) and several of the ocaml-ci builds are expected to fail due to static linking issues on some distros I think we're good.

LGTM!

PR #647, opened 2025-04-03
Add GHA workflow to build image

vvoland:
Based off: https://github.com/moby/moby/blob/master/.github/workflows/bin-image.yml

### DIFF
--- a/.github/workflows/bin-image.yml
+++ b/.github/workflows/bin-image.yml
@@ -25,160 +25,18 @@ on:
 env:
   REPO_SLUG: moby/vpnkit-bin
 jobs:
-  prepare:
-    runs-on: ubuntu-24.04
-    timeout-minutes: 20
-    outputs:
-      platforms: ${{ steps.platforms.outputs.matrix }}
-    steps:
-      -
-        name: Checkout
-        uses: actions/checkout@v4
-      -
-        name: Docker meta
-        id: meta
-        uses: docker/metadata-action@v5
-        with:
-          images: |
-            ${{ env.REPO_SLUG }}
-          tags: |
-            type=semver,pattern={{version}}
-            type=ref,event=branch
-            type=ref,event=pr
-            type=sha
-      -
-        name: Rename meta bake definition file
-        # see https://github.com/docker/metadata-action/issues/381#issuecomment-1918607161
-        run: |
-          bakeFile="${{ steps.meta.outputs.bake-file }}"
-          mv "${bakeFile#cwd://}" "/tmp/bake-meta.json"
-      -
-        name: Upload meta bake definition
-        uses: actions/upload-artifact@v4
-        with:
-          name: bake-meta
-          path: /tmp/bake-meta.json
-          if-no-files-found: error
-          retention-days: 1
-      -
-        name: Create platforms matrix
-        id: platforms
-        run: |
-          echo "matrix=$(docker buildx bake bin-image --print | jq -cr '.target."bin-image".platforms')" >>${GITHUB_OUTPUT}
-
   build:
-    runs-on: ${{ contains(matrix.platform, 'arm') && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
-    timeout-minutes: 45 # guardrails timeout for the whole job
-    needs:
-      - prepare
-    if: always() && !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled')
-    strategy:
-      fail-fast: false
-      matrix:
-        platform: ${{ fromJson(needs.prepare.outputs.platforms) }}
-    steps:
-      -
-        name: Checkout
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-      -
-        name: Prepare
-        run: |
-          platform=${{ matrix.platform }}
-          echo "PLATFORM_PAIR=${platform//\//-}" >> $GITHUB_ENV
-      -
-        name: Download meta bake definition
-        uses: actions/download-artifact@v4
-        with:
-          name: bake-meta
-          path: /tmp
-      -
-        name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
-      -
-        name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
-        with:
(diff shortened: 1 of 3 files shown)


vvoland:
Workflow won't run in this PR before merging. Workflow tested in other repo: https://github.com/docker/vvoland-gha-tests/actions/runs/14308322398

.github/workflows/bin-image.yml
```diff
@@ -0,0 +1,184 @@
...
+    tags:
+      - 'v*'
+  pull_request:
+
+env:
+  REPO_SLUG: moby/vpnkit-bin
+jobs:
+  prepare:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 20 # guardrails timeout for the whole job
```
thaJeztah, line 30: Is 20 minutes enough for this one? I recall you mentioned building the image here was rather slow

vvoland: It's only the prepare step. Build has bigger timeout:
https://github.com/moby/vpnkit/blob/207eb9fbe7d9b74d6e05f1481cca0ca6cfcc9c8a/.github/workflows/bin-image.yml#L71

OTOH, it could even be lower (like 60 minutes).

The workflow is much faster now (it took ~10 minutes in the private repo) due to using the native github runner:

https://github.com/moby/vpnkit/blob/207eb9fbe7d9b74d6e05f1481cca0ca6cfcc9c8a/.github/workflows/bin-image.yml#L70

thaJeztah:
I tried if I could modernise the Dockerfile a bit, and to add build-args to allow overriding the versions (also getting rid of that hard-coded `/home/opam/.opam/4.14/bin/vpnkit` path, but ran into issues with permissions because the image runs as `opam`, which didn't have permissions to create the output dir 😂 

For posterity; this is what I had;


```Dockerfile
# syntax=docker/dockerfile:1

ARG ALPINE_VERSION=3.19
ARG OCAML_VERSION=4.14
ARG OCAML_TAG=alpine-${ALPINE_VERSION}-ocaml-${OCAML_VERSION}

FROM ocaml/opam:${OCAML_TAG} AS build
RUN opam update

COPY --link . /home/opam/vpnkit
RUN opam pin add vpnkit /home/opam/vpnkit --kind=path -n
RUN opam depext vpnkit -y
RUN opam install vpnkit -y --destdir /home/opam/build

FROM scratch AS binary
COPY --link --from=build /home/opam/build/bin/vpnkit /vpnkit

FROM alpine:latest
COPY --link --from=binary /vpnkit /vpnkit
```

.github/workflows/bin-image.yml
```diff
@@ -0,0 +1,184 @@
...
+        working-directory: /tmp/digests
+        run: |
+          set -x
+          docker buildx imagetools create $(jq -cr '.target."docker-metadata-action".tags | map("-t " + .) | join(" ")' /tmp/bake-meta.json) \
+            $(printf '${{ env.REPO_SLUG }}@sha256:%s ' *)
+      -
+        name: Inspect image
+        run: |
+          set -x
+          docker buildx imagetools inspect ${{ env.REPO_SLUG }}:$(jq -cr '.target."docker-metadata-action".args.DOCKER_META_VERSION' /tmp/bake-meta.json)
```
crazy-max, line 184: I think we could make use of this reusable workflow https://github.com/crazy-max/.github?tab=readme-ov-file#build-distribute-mp:

```suggestion
  build:
    uses: crazy-max/.github/.github/workflows/build-distribute-mp.yml@d9a10e2737504a6e253f96e344cef684b0e00cb5
    with:
      target: bin-image
      push: ${{ github.event_name != 'pull_request' && github.repository == 'moby/vpnkit' }}
      cache: true
      meta-image: moby/vpnkit
      meta-tags: |
        type=semver,pattern={{version}}
        type=ref,event=branch
        type=ref,event=pr
        type=sha
      build-platforms: linux/amd64,linux/arm64
    secrets:
      login-username: ${{ secrets.DOCKERHUB_USERNAME }}
      login-password: ${{ secrets.DOCKERHUB_TOKEN }}
```

Reminds me I need to update https://github.com/moby/moby/pull/49474

vvoland: That won't allow to choose a correct gha runner though?

https://github.com/moby/vpnkit/blob/3d4f258e7514b9cc878639f724cbb0caffa8fd98/.github/workflows/bin-image.yml#L70

It's kinda important here because otherwise the build takes ~30 minutes instead of ~5 😅

crazy-max: > Ah, actually it does handle it too (`runner` defaults to `auto`), nice!

Yes indeed `auto` choose based on target platform

vvoland: Updated to use your workflow, I did set `cache` to `false` though.

crazy-max: > Looks like the warmup stage only builds one platform: https://github.com/docker/vvoland-gha-tests/actions/runs/14330651343/job/40165666013
> 
> Effectively this turns a parallel build into sequential as the linux/arm64 doesn't run until the warmup stage finishes (which caches the linux/amd64).
> 
> Is there a way to disable the warmup without disabling the gha cache completely?

I can add an option for that, will take a look

crazy-max (approved):

.github/workflows/bin-image.yml
```diff
@@ -0,0 +1,42 @@
...
+env:
+  REPO_SLUG: moby/vpnkit-bin
+jobs:
+  build:
+    uses: crazy-max/.github/.github/workflows/bake-distribute-mp.yml@d9a10e2737504a6e253f96e344cef684b0e00cb5
+    with:
+      target: bin-image
+      push: ${{ github.event_name != 'pull_request' && github.repository == 'moby/vpnkit' }}
+      cache: false # See: https://github.com/moby/vpnkit/pull/647/files/3d4f258e7514b9cc878639f724cbb0caffa8fd98#r2032880337
+      meta-image: moby/vpnkit
```
crazy-max, line 34: Ah looking at https://github.com/moby/vpnkit/actions/runs/14331701104/job/40169030245#step:7:764 I think it should be `moby/vpnkit-bin`

PR #645, opened 2024-11-19
Correctly return NOERROR even if host resolver returned empty list

dan0dbfe:
Fixes #509 and https://github.com/moby/moby/issues/47628.

1. `Dns.resolve` now returns  `(rr list, unit) result` to distinguish between `[]` resulting from NOERROR or NXDOMAIN.
    - Currently only returns NXDOMAIN on any kind of error.
    - `Dns.getaddrinfo` returns `(rr list, Msg) result` with no error handling just like current behavior
2. `Make.answer` returns answers (including `[]`) if there is no error, otherwise NXDOMAIN is returned.
3. Added NOERROR and NXDOMAIN test cases.

Issue #509, opened 2020-08-31
VPNKit DNS server returns NXDOMAIN for SRV records


AkihiroSuda:
VPNKit DNS server returns NXDOMAIN for SRV records
```console
$ rootlesskit --net=vpnkit dig -t srv _imaps._tcp.gmail.com
WARN[0000] specifying --disable-host-loopback is highly recommended to prohibit connecting to 127.0.0.1:* on the host namespace (requires slirp4netns or VPNKit) 
WARN[0000] Mounting /etc/resolv.conf without copying-up /etc. Note that /etc/resolv.conf in the namespace will be unmounted when it is recreated on the host. Unless /etc/resolv.conf is statically configured, copying-up /etc is highly recommended. Please refer to RootlessKit documentation for further information. 

; <<>> DiG 9.16.1-Ubuntu <<>> -t srv _imaps._tcp.gmail.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 870
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;_imaps._tcp.gmail.com.         IN      SRV

;; Query time: 0 msec
;; SERVER: 192.168.65.1#53(192.168.65.1)
;; WHEN: Mon Aug 31 20:02:57 JST 2020
;; MSG SIZE  rcvd: 39
```

OTOH slirp4netns DNS works as expected:
```console
$ ./rootlesskit --net=slirp4netns dig -t srv _imaps._tcp.gmail.com
WARN[0000] specifying --disable-host-loopback is highly recommended to prohibit connecting to 127.0.0.1:* on the host namespace (requires slirp4netns or VPNKit) 
WARN[0000] Mounting /etc/resolv.conf without copying-up /etc. Note that /etc/resolv.conf in the namespace will be unmounted when it is recreated on the host. Unless /etc/resolv.conf is statically configured, copying-up /etc is highly recommended. Please refer to RootlessKit documentation for further information. 

; <<>> DiG 9.16.1-Ubuntu <<>> -t srv _imaps._tcp.gmail.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 34903
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

[...]

PR #649, opened 2025-06-26
Fix(DNS): Handle EAI_NODATA as success (empty address list) in getaddrinfo

keigoi:
This pull request addresses a long-standing issue where DNS lookups within containers, especially those running in rootless Docker-in-Docker (DIND) environments using VPNKit, would fail with an `NXDOMAIN` error. 

This often occurred even when valid IP addresses were present in the DNS response, or for specific record types like SRV records.

#### Problem:

*   **DNS Lookup Errors in Rootless DIND**
    *   Related Issue: moby/moby#47628
    *   When running containers in `dind-rootless` mode, DNS lookups would fail with `NXDOMAIN`.
    *   This was particularly evident when the upstream DNS server stripped out IPv6 addresses. For example, commands like `apk add` in `alpine` containers would fail due to these DNS errors.
    *   This behavior was reproduced across multiple dind images.

*   **NXDOMAIN for SRV Records**
    *   The VPNKit DNS server was specifically observed to return `NXDOMAIN` for SRV record queries, unlike `slirp4netns` which handled them correctly (https://github.com/moby/vpnkit/issues/509).

The core of the issue stemmed from VPNKit treating the `EAI_NODATA` error code from `getaddrinfo` as a fatal failure. `EAI_NODATA` indicates that the name exists, but there are no addresses of the requested type (e.g., no AAAA records if IPv6 is filtered out, or no SRV records for a specific query).

#### Solution:

This pull request modifies VPNKit's DNS handling to **interpret `EAI_NODATA` as a successful result, specifically as an empty address list**. By doing so, DNS lookups can complete successfully even when an upstream DNS server returns an empty list for a particular query type.

#### Expected Outcome:

*   DNS lookups within `rootless DIND` containers using VPNKit will now work as expected, even when upstream DNS servers strip IPv6 addresses.
*   VPNKit-based SRV record lookups will no longer return `NXDOMAIN` when the name exists but no SRV records are found, allowing for proper resolution or indicating an empty SRV record list.
*   This change directly resolves the issues reported in moby/moby#47628 and moby/vpnkit#509.

#### Notes

* While Pull Request #645 aimed to fix the issue, it did not fully address the fundamental problem of how VPNKit handled EAI_NODATA error codes from getaddrinfo.

#### Acknowledgments

* Special thanks to @joanbm, who is explicitly listed as a co-author on this commit (the fullname and email address is retrieved from their own public repo). 
    *  See: https://github.com/moby/vpnkit/issues/509#issuecomment-2892480710

* A special thanks to Tomoya Kawaguchi (@yamoyamoto) for their invaluable help in debugging this issue. Tomoya added debug logs to narrow down the problem and confirmed that nslookup no longer returned NXDOMAIN within the Alpine images in the DIND environment.

### DIFF
--- a/src/hostnet/host.ml
+++ b/src/hostnet/host.ml
@@ -1267,6 +1267,10 @@ module Dns = struct
   let getaddrinfo node family =
     Luv_lwt.in_luv (fun return ->
         Luv.DNS.getaddrinfo ~family ~node () (function
+          | Error `EAI_NODATA ->
+              (* Special handling for EAI_NODATA: Treat as success (host exists, but no address data) and
+                 Return an empty list. See https://github.com/moby/moby/issues/47628 for more context. *)
+              return (Ok [])
           | Error err -> return (Error (`Msg (Luv.Error.strerror err)))
           | Ok x ->
               let ips =


dan0dbfe:
Looks like something changed that broke the dep solver in the build stage of `ocaml-ci`.

I'm guessing the package upgrades are unintentional but a side effect of opam-repository updating something.

[In the last PR that worked](https://ocaml.ci.dev/github/moby/vpnkit/commit/decfffb34043f0543c8564a289d6e72c8834322d/variant/%28analysis%29?show_raw=true):
> Solving with opam-repository commit: https://github.com/ocaml/opam-repository.git#refs/heads/master (ad2202b486885c0c13f146ead5c85646ce87e24e)

and it used `cmdliner.1.0.4` for example.

However [this PR's run](https://ocaml.ci.dev/github/moby/vpnkit/commit/13fb1c5e679ede1fe603cec14741aa84768c7a90/variant/%28analysis%29?show_raw=true) failed. The commit used was:

> Solving with opam-repository commit: https://github.com/ocaml/opam-repository.git#refs/heads/master (3e4a334c6caed0798833ef331ef815eede191f03)

and it failed to solve as, amongst other problems, `alcotest 1.9.0 requires >= 1.2.0`

PR #653, opened 2025-11-14
Github action to build with dune package management

gridbugs:
This adds a dune lockdir to the project and a github action where pushing a tag causes a vpn binary distribution to be built for linux and macos, and released to the github "releases" page.

Also fixes a few minor issues which prevented the project from building with the locked versions of dependencies.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,42 @@
+name: release
+
+on:
+  push:
+    tags:
+    - '*'
+
+jobs:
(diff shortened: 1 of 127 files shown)


gridbugs:
> Nice! Any idea if this could also work for building a windows binary (ideally statically linked to help with distribution)?

There's some more work needed in Dune to better support windows. Currently certain key packages including the OCaml compiler don't build on windows with package management. See https://github.com/ocaml/dune/issues/11161

PR #657, opened 2026-03-23
Remove `REPO_ROOT` variable and `depends.sh`

Leonidas-from-XIV:
The `REPO_ROOT` variable from the Makefile does not seem to be used anywhere. The only mention of `REPO_ROOT` is `scripts/depends.sh` which is also not referenced anywhere (and not called in the `Makefile` either).

This PR removes both of them.


Leonidas-from-XIV:
@djs55 Could you please take a look? This removes a dependency on a `git` call that seems unnecessary overall.

Leonidas-from-XIV:
@djs55 Do you have a moment to review this?

Leonidas-from-XIV:
@djs55 Do you know if there's a specific use for this code? It didn't seem that way to me unless I missed something and removing it saves a `git` call.

djs55 (approved):

PR #660, opened 2026-03-23
Use `public_name` to install public binary

Leonidas-from-XIV:
The dune files build `main` and then rename it to `vpnkit` and install it. This what `public_name` does, thus this PR removes the copy logic and the install customization and adds a `public_name` stanza instead.

### DIFF
--- a/dune
+++ b/dune
@@ -17,11 +17,6 @@
  (action (run %{gen} -out %{target} -in %{deps})))
 
 (rule
- (target vpnkit.exe)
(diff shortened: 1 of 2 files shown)


Leonidas-from-XIV:
@djs55 Can you take a look at this PR? It might be that the way it was done before had some particular reason but from the top of my head it seems like using `public_name` is a better option.

djs55 (approved):

